PR #9, opened 2022-04-09
howto for OMO validation

cmungall:


matentzn:
How does this fit with with Jims SHACL work? I would prefer it if I could load an ontology natively (without need to sqlite conversion) and run --validate shape.shacl, so we don't need to re-invent yet another validation language.. James also has an entire library for validation based as well.. Sorry don't want to spoil the fun, but feels like we are fragmenting too much :/

cmungall:
I wasn’t aware there was much activity on the shacl side!

Re SQLite dependency. That is just one implementation. Other
implementations possible. I would bet Postgres the fastest for validating
all of obo.

cmungall:
See: https://github.com/cmungall/obo-metadata

PR #17, opened 2022-04-14
Add get_sssom_mappings_by_curie to BioportalImplementation class

pkalita-lbl:
This method uses the [`/ontologies/:ontology/classes/:cls/mappings` endpoint](https://data.bioontology.org/documentation#Mapping) to get mappings and reformats them in SSSOM format.

Caveats:
* The BioPortal API has an [issue](https://github.com/ncbo/ontologies_linked_data/issues/117) where a lot of duplicate mappings are reported. We are not trying to resolve that duplication here and instead just passing the results through.
* The BioPortal API reports a `source` field for each mapping and we are manually translating that to a SKOS term, but this may be up for refinement
* The method will accept a CURIE as defined by the `MappingProviderInterface` but since the BioPortal API generally deals with URIs and not CURIEs the SSSOM `subject_id` and `predicate_id` fields will be URIs

### DIFF
--- a/src/oaklib/implementations/bioportal/bioportal_implementation.py
+++ b/src/oaklib/implementations/bioportal/bioportal_implementation.py
@@ -1,22 +1,36 @@
-import requests
 import logging
 from dataclasses import dataclass, field
-from typing import Any, List, Dict, Union, Iterator, Iterable, Tuple
+from typing import Any, Dict, Iterable, Iterator, List, Tuple, Union
+from urllib.parse import quote
 
+import requests
 from oaklib.datamodels.text_annotator import TextAnnotation
 from oaklib.interfaces.basic_ontology_interface import PREFIX_MAP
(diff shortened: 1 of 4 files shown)


src/oaklib/implementations/bioportal/bioportal_implementation.py
```diff
@@ -1,22 +1,36 @@
...
 
 ANNOTATION = Dict[str, Any]
 
+# See: 
+#   https://www.bioontology.org/wiki/BioPortal_Mappings 
+#   https://github.com/agroportal/project-management/wiki/Mappings
+SOURCE_TO_PREDICATE = {
+    'CUI': 'skos:closeMatch',
+    'LOOM': 'skos:closeMatch',
+    'REST': 'skos:relatedMatch', # maybe??
```
cmungall, line 27: yeh it's not super clear which skos property to use (other than SAME_URI), I think this is good enough to get started!

cmungall (approved):

PR #21, opened 2022-04-18
Sphinx documentation

### LINKED ISSUE #14: Add gh action to publish docs to gh-pages
Use
https://github.com/marketplace/actions/sphinx-to-github-pages

@hrshdhgd - I think we should do this for other repos too

hrshdhgd:
Fixes #14 

This is how this works:

The documentation files will be hosted in the `gh-pages` branch and it will be deployed only when a PR is merged to the main branch. The `docs` folder is no longer needed in the main branch anymore.


cmungall (changes requested):
the source (anything authored) should remain on main

derived html should go on gh pages

hrshdhgd:
Flawed logic and workflow... restarting

PR #23, opened 2022-04-19
poetry update + tested workflow for sphinx deployment in gh-pages

### LINKED ISSUE #14: Add gh action to publish docs to gh-pages

hrshdhgd:
Fixes #14 

This is how this works:

The documentation files will be hosted in the `gh-pages` branch and it will be deployed only when a PR is merged to the main branch. The `docs` folder  in the main branch hosts all the `*.rst` files needed by sphinx to generate the documentation.

### DIFF
--- a/.github/workflows/doc_pages.yaml
+++ b/.github/workflows/doc_pages.yaml
@@ -1,0 +1,40 @@
+name: Sphinx Documentation
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  build-docs:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@main
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/doc_pages.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+    - name: install
+      run: poetry install -E docs
+
+    - name: Build documentation.
+      run: |
+        mkdir gh-pages
+        touch gh-pages/.nojekyll
+        cd docs/
+        poetry run sphinx-build -b html . _build
```
hrshdhgd, line 31: This step shows that the `docs` folder in the main branch is required.

cmungall (approved):

PR #48, opened 2022-04-28
test

cmungall:


cmungall:
see also https://github.com/hsolbrig/dataclasstest

cmungall:
works (3.10 fails as expected)

PR #56, opened 2022-05-02
Updates to BioportalImplementation to support use in mapping-walker

pkalita-lbl:
To use `BioportalImplementation` in mapping-walker it needs a method to get the ancestors of a term. In order to make a request to the BioPortal API, an ontology acronym (e.g. `FMA`) and a class IRI ID (e.g. `http://purl.org/sig/ont/fma/fma63179`) are needed. As far as I can tell there isn't a way to derive one from the other, so these changes also add a cache of class-to-ontology associations that is populated on the call to `get_sssom_mappings_by_curie`. This means that there is an unstated assumption that to get "good" results, `get_sssom_mappings_by_curie` will be called before `ancestors` (as mapping-walker does).


pkalita-lbl:
I did something weird when trying to merge main into this branch. Closing this PR and will try again.

PR #70, opened 2022-05-05
Issue 40

### LINKED ISSUE #40: Add a notebook demonstrating mappings
in the notebooks/ folder

rough script of ow it should work:

part 1 - fetching existing mappings

extract a sssom file from bioportal
extract a sssom file from a downloaded ontology

choose an uberon. term like limb as an example

part 2 - generating mappings

choose two small ontologies, e.g. zfa and xao, demonstrate using lexmatch to find mappings

hrshdhgd:
Fixes #40


hrshdhgd:
Here's the [notebook](https://nbviewer.org/github/INCATools/ontology-access-kit/blob/issue_40/notebooks/mappings.ipynb)

PR #125, opened 2022-06-17
[WIP] Pass flake8

cthoyt:
This PR turns on `flake8` in earnest. It adds two ignores - E501 (line too long) and W503 (line break after operator). These are okay since flake8 is actually wrong about W503 and E501 is taken care of by `black`. Additional ignores can be used to reduce the scope of this PR, if desired, but eventually it makes sense to pass flake8 in its entirety.

PR #131, opened 2022-06-22
More flake8

cthoyt:
Follow-up to #125


cthoyt:
time for round 2 @hrshdhgd @sujaypatil96 @cmungall - this time fixing the remaining flake8 issues. Looking forward, round 3 will begin introducing more plugins to find even further code issues

sierra-moxon:
@cthoyt - why not just branch of main for these?  You have permission right?

hrshdhgd:
This is done. I have added 2 flake8 errors in the `ignore` list for now to tackle in a separate PR. 

```
C901 
E731
```

If we could continue with flake8 edits in the future through branches within the same org 'INCATools', it would be a little convenient to work with.

PR #127, opened 2022-06-17
Add `class-resolver`-based materialization of interface objects

cthoyt:
Closes #126 

- [x] implements [`class-resolver`](https://github.com/cthoyt/class-resolver) lookup
- [x] Adds `class-resolver` as dependency (idk how poetry works and there's no contribution guide in this repo yet :p)

## Example usage

```python

from oaklib import OntologyResource

ontology_resource = OntologyResource(slug='tests/input/go-nucleus.db', local=True)

# Defaults to SQL implementation
ontology_interface = ontology_resource.materialize()

# Is smart and figures out which is the SQL one by name
# i.e., it knows to append "Implementation" at the end of the name
ontology_interface = ontology_resource.materialize("sql")

# Explicit
from oaklib.implementations import SqlImplementation
ontology_interface = ontology_resource.materialize(SqlImplementation)

# Forwards kwargs to SqlImplementation or whatever implementation you picked
from sqlalchemy import create_engine
engine = create_engine('postgresql://scott:tiger@localhost:5432/mydatabase')
ontology_interface = ontology_resource.materialize("sql", engine=engine)
```

PR #138, opened 2022-06-23
Addressing C901 and E731

hrshdhgd:
Follow-up to #131 
Addressing https://github.com/INCATools/ontology-access-kit/pull/131#issuecomment-1164755651

 - C901 - Function is too complex
 - E731 - Do not assign a lambda expression, use a def

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -37,8 +37,6 @@ description = Run the flake8 tool with several plugins (bandit, docstrings, impo
 ignore =
     E203
     W503
-    C901 # needs code change so ignoring for now.
-    E731 # needs code change so ignoring for now.
 max-line-length = 120
 max-complexity = 13
 import-order-style = pycharm


cmungall:
this seems to involve some changes that in selectors that are nothing to do with flake8?

src/oaklib/selector.py
```diff
@@ -38,6 +38,22 @@
     "json-ld": "json-ld",
 }
 
+SCHEME_DICT = {
```
cthoyt, line 41: good chance to look into #127, which uses `class-resolver` to accomplish things like this

cthoyt:
@hrshdhgd would be great to get this finished up - are there any blockers?

cthoyt:
I think we should still finish this PR and merge it in as a stop-gap. Most importantly, this PR modularizes the logic for actually picking the implementation, and in #127 as a follow-up we can replace the dictionary itself.

hrshdhgd:
Well, in that case, the checks pass, I don't see any pending work on this PR.

Issue #136, opened 2022-06-23
where does the sqlite database go when I do something like --input sqlite:obo:envo


turbomam:

cmungall:
Yes, this needs to be better documented. 

API keys, caches, etc are all controlled by

https://pypi.org/project/appdirs/

Which chooses an OS-appropriate location.

You can increase `verbosity` to see what is going on:

```
✗ runoak -v -i sqlite:obo:obi info ...
INFO:root:RESOURCE=OntologyResource(slug='obo:obi', directory=None, scheme='sqlite', format=None, url=None, readonly=False, provider=None, local=False, in_memory=False, data=None, implementation_class=<class 'oaklib.implementations.sqldb.sql_implementation.SqlImplementation'>)
INFO:root:Locator: obo:obi
INFO:root:Using cache dir: /Users/cjm/Library/Caches/oaklib
INFO:root:Downloading from https://s3.amazonaws.com/bbop-sqlite/obi.db to /Users/cjm/Library/Caches/oaklib/obi.db
```

FAQ PRs welcome! Or in the docs here:

https://incatools.github.io/ontology-access-kit/intro/tutorial07.html

But we should decide whether we want to keep using appdirs or @cthoyt's pystow

See:

- #110

cthoyt:
I'm biased but [`PyStow`](https://github.com/cthoyt/pystow) has all of the functionality I always wished for when it comes to download and caching. It has options to use the XDG-standard locations, but I found that putting a dot directory in the home directory is actually much more transparent (which pystow does by default)

turbomam:
Thanks. `/Users/MAM/Library/Caches/oaklib/envo.db` in MacOS 12.4

PR #140, opened 2022-06-24
Demonstrate using PyStow for downloading S3 SQL content

cthoyt:
Related to #136 

PyStow has all of the logic of caching files from a URL (including deterministically naming the files) built in. It can be used to replace a small amount of code that works with S3 to get pre-build SQLite versions of ontologies.

- [x] Replaces implementation of `download()`
- [x] Replaces downloading logic with PyStow's `ensure()` functionality 
- [x] Documents how PyStow chooses its caching location (e.g., it puts stuff in `~/.data/oaklib` but this can be modified with certain environment variables to either use XDG-compliant paths or to change `.data` to something else explicitly)

I've also given a second option in the comments that uses PyStow's S3 API wrapper (via `botocore`)


src/oaklib/implementations/sqldb/sql_implementation.py
```diff
@@ -95,6 +90,12 @@
 from oaklib.interfaces.validator_interface import ValidatorInterface
 from oaklib.types import CURIE, SUBSET_CURIE
 
+__all__ = [
```
cmungall, line 93: Aside: Do you think we should implement this pattern throughout

(when looking at this I wonder if these should actually be not exported and start with `_` but can handle that separately)

src/oaklib/implementations/sqldb/sql_implementation.py
```diff
@@ -111,14 +112,6 @@ def _mapping(m: sssom.Mapping):
     return m
 
 
-# https://stackoverflow.com/questions/16694907/download-large-file-in-python-with-requests
```
cmungall, line 114: glad to see this go.

Note to self: we have gz files on s3 now will want to download these preferentially

cmungall (approved):

cmungall:
We discussed this a bit externally, there was some concern about introducing an additional dependency for this task, but I pointed out we are already using appdirs

There was also some concern about hardwiring to S3, but that had nothing to do with this PR (the option to use S3-specific is commented out), I made a separate issue to consider PURLs for the sqlite downloads: https://github.com/INCATools/semantic-sql/issues/43

src/oaklib/implementations/sqldb/sql_implementation.py
cthoyt: Yes, we should carefully do exports everywhere.

PR #141, opened 2022-06-24
Add flake8 plugins: bandit, bugbear, pep8-naming

cthoyt:
Time for some of the heavy flake8 pulls - this PR introduces bandit (security issues), bugbear (finding code smells like unused iteration values), and pep8-naming (this is going to yell about bad variable names)


cthoyt:
@hrshdhgd also consider that every PR to the repo where these checks aren't operationalized will make doing this more difficult - technical debt piles up very fast

hrshdhgd:
I've set the following to the ignore list for now:

```
    B008 # Do not perform function calls in argument defaults.
    S404 # Consider possible security implications associated with the subprocess module.
    S108 # Probable insecure usage of temp file/directory.
    S307 # Use of possibly insecure function - consider using safer ast.literal_eval.
    S603 # subprocess call - check for execution of untrusted input.
    S607 # Starting a process with a partial executable path ["open" in both cases]
    S608 # Possible SQL injection vector through string-based query construction.
```

cmungall (approved):

PR #148, opened 2022-06-30
refactor query term iterator

cmungall:
- refactored query iterator
- Adding nesting of search terms

### DIFF
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -25,6 +25,7 @@ from typing import (
     List,
     Optional,
     TextIO,
+    Tuple,
     Type,
     Union,
 )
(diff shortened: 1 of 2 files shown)


src/oaklib/cli.py
```diff
@@ -202,13 +204,59 @@ def _shorthand_to_pred_curie(shorthand: str) -> PRED_CURIE:
...
+    if rest:
+        raise ValueError(f"Unparsed: {rest}")
+    return nested
+
+
+def _nest_list_of_terms(terms: List[str]) -> Tuple[NESTED_LIST, List[str]]:
+    nested = []
+    while len(terms) > 0:
+        term = terms[0]
+        terms = terms[1:]
```
hrshdhgd, line 231: minor but can't lines 230 and 231 be replaced by `term = terms.pop(0)`

src/oaklib/cli.py
```diff
@@ -202,13 +204,59 @@ def _shorthand_to_pred_curie(shorthand: str) -> PRED_CURIE:
...
+        if term == "[":
+            nxt, rest = _nest_list_of_terms(terms)
+            terms = rest
+            nested.append(nxt)
+        elif term == "]":
+            return nested, terms
+        else:
+            nested.append(term)
+    return nested, []
```
hrshdhgd, line 240: From here I get a sense that there is a possibility of nested lists. Would there be a disadvantage if the nested list was flattened first? 

`terms = list(itertools.chain(*terms))` or is it required to preserve the nesting in the list?

src/oaklib/cli.py
cthoyt:
```
term = terms[0]
terms = terms[1:]
```

can be replaced with clever usage of the splat operator

```
term, *terms = terms
```

src/oaklib/cli.py
cthoyt: consider also `list(itertools.chain.from_iterable(terms))`

cmungall: this is intentionally nested. There is an experimental feature where you can nest the boolean combinations of search directives. This will likely migrate to a full query datamodel rather than overloading lists

hrshdhgd (approved):

PR #152, opened 2022-06-30
Address pending flake8 rules

hrshdhgd:
Continued from #141 
 - [x] B008 # Do not perform function calls in argument defaults.

### DIFF
--- a/src/oaklib/implementations/ols/__init__.py
+++ b/src/oaklib/implementations/ols/__init__.py
@@ -1,0 +1,3 @@
+from oaklib.datamodels.search import SearchConfiguration
+
+SEARCH_CONFIG = SearchConfiguration()
--- a/src/oaklib/implementations/bioportal/__init__.py
+++ b/src/oaklib/implementations/bioportal/__init__.py
@@ -1,0 +1,3 @@
+from oaklib.datamodels.search import SearchConfiguration
+
+SEARCH_CONFIG = SearchConfiguration()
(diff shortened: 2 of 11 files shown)


cmungall:
what's the SQL issue? We should be using SQLA everywhere but there may be a couple of holdover direct accesses

hrshdhgd:
> what's the SQL issue? We should be using SQLA everywhere but there may be a couple of holdover direct accesses

```
rows = list(con.execute(f"SELECT * FROM {TBL_NAME}"))
```
Possible SQL injection vector through string-based query construction.

hrshdhgd:
For` S307 # Use of possibly insecure function - consider using safer ast.literal_eval.`:

`eval` seems to yield results whereas `ast.literal_eval` gives an error stating: 
```
*** TypeError: literal_eval() takes 1 positional argument but 2 were given
```

So I will skip it until someone can come up with a different solution

cmungall:
Ah, here:

https://github.com/INCATools/ontology-access-kit/blob/fb6b3d55da0073dd2f857c3c2ba78582ce51cee7/tests/test_utilities/test_sqlite_utils.py#L16

That code should be moved from the repo anyway and put in sem-sql

cmungall:
How should we progress on this one @hrshdhgd?

If you would like to check a few more boxes, let's mark this as draft

But I think it better to do this incrementally - in which case maybe edit the first comment so that the checklist is split into the items to be done for this PR vs ones saved for another

hrshdhgd:
I could move the unchecked ones to a new issue and merge this PR with just the one checked rule addressed. Does that sound reasonable?

cmungall:
Sounds good!

src/oaklib/implementations/ols/__init__.py
```diff
@@ -0,0 +1,3 @@
+from oaklib.datamodels.search import SearchConfiguration
+
+SEARCH_CONFIG = SearchConfiguration()
```
cthoyt, line 3: this is pretty sketchy, why are there several instances of this objects in different static modules? are these mutable? if so, they shouldn't be the default values for functions regardless of if they live in a module or are constructed in the specification of a function

PR #157, opened 2022-07-04
Documentation for dev/setup (old)

joeflack4:
- Updates: README.md: (i) reorganized/added 'dev' section, (ii) added project setup steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,17 @@ Same using pronto, fetching ontology from obolibrary
 runoak -i obolibrary:go.obo  viz GO:0005773
 ```
 
-## Potential Refactoring
+## Developer notes
+### Local project setup
+Prerequisites:
+1. Python 3.10+
+2. [Poetry](https://python-poetry.org/)
+
+Setup steps:
+1. Clone this repository.
+2. From the repo directory, run: `poetry install`
+
+### Potential Refactoring
 
 Currently all implementations exist in this repo/module, this results in a lot of dependencies
 


README.md
```diff
@@ -215,7 +215,17 @@ Same using pronto, fetching ontology from obolibrary
 runoak -i obolibrary:go.obo  viz GO:0005773
 ```
 
-## Potential Refactoring
+## Developer notes
+### Local project setup
+Prerequisites:
+1. Python 3.10+
```
cmungall, line 221: 3.9 works

In fact if you look at the checks we do
https://github.com/INCATools/ontology-access-kit/blob/main/.github/workflows/main.yaml

The *only* guarantee we have is for 3.9

joeflack4: Thanks! This was just a guess on my part. will change it now

joeflack4:
Looks like I still can't push to this trunk repository directly. I was looking at the [team permissions](https://github.com/orgs/INCATools/teams/oak-developers/repositories), and it looks like it only gives 'Read' access..

README.md
```diff
@@ -215,7 +215,17 @@ Same using pronto, fetching ontology from obolibrary
...
+2. [Poetry](https://python-poetry.org/)
+
+Setup steps:
+1. Clone this repository.
```
cthoyt, line 225: would be better to explicitly write the commands in a shell-like format, including the git clone command, the cd, etc

See https://github.com/biopragmatics/bioregistry#-installation for an example

joeflack4: Not a bad suggestion. Done~

I did leave off the `$` though. Even though it's a common convention, I have some beef with that: (i) makes copy/paste more tedious, (ii) the code block formatting already helps with indication, (iii) not every prompt uses `$`.

joeflack4:
Replacing this with #162

PR #159, opened 2022-07-06
Add text annotator interface via Gilda

cthoyt:
This PR adds a text annotator interface for Gilda, which only does named entity normalization (ground) and explicitly _not_ named entity recognition, but the default implementation for this class makes it look like that should be okay.

More information about Gilda can be found on the GitHub repository at https://github.com/indralab/gilda and the accompanying paper:

> [Gilda: biomedical entity text normalization with machine-learned disambiguation as a service](https://doi.org/10.1093/bioadv/vbac034)
Benjamin M Gyori, Charles Tapley Hoyt, Albert Steppi (2022)
*Bioinformatics Advances*, vbac034


cthoyt:
A few follow-up questions from me and @bgyori:

1. Should the annotator yield all possible groundings for the same string?
2. How should we handle grounding to non ontology terms (e.g. UniProt)?
3. How do we annotate the requirement for `gilda`?

src/oaklib/implementations/gilda.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def annotate_text(
+        self, text: TEXT, configuration: TextAnnotationConfiguration = None
+    ) -> Iterator[TextAnnotation]:
+        if not configuration:
+            raise NotImplementedError
+        if not configuration.matches_whole_text:
+            raise NotImplementedError
+
+        import gilda
```
cmungall, line 33: Wondering about this. How would this work when we want to add a unit test for this?

My preference would be either:

1. have a separate gilda plugin (preliminary docs here file:///Users/cjm/repos/ontology-access-kit/docs/_build/html/howtos/write-a-plugin.html)
2. add this to pyproject.toml + poetry.lock (maybe that is the ultimate the intent and you just didn't want to risk commit conflicts for now?)

src/oaklib/implementations/gilda.py
```diff
@@ -0,0 +1,39 @@
...
+
+        matches = gilda.ground(text)
```
cmungall, line 35: any thoughts on how to incorporate the context?

(can be done on a separate PR)

I can see two ways

1. a separate context argument in the interface (that many implementations can choose to ignore)
2. treat the full text span as the context, and instead add an optional start/end to the call to the interface, essentially allowing the caller to say "annotate this region of the text using the whole text as context"

2 is maybe a bit more fiddly for the client

cmungall (commented):
lmk your thoughts on the package dependency issue, happy to do this incrementally though

src/oaklib/implementations/gilda.py
cthoyt: Gilda supports the context argument out of the box but getting into that calls bigger questions about doing NER vs NEN with the same Interface

cthoyt: I’d say just passing it as an extra argument would me much more simple than fiddling around with text positions

src/oaklib/implementations/gilda.py
cthoyt: I really want to stress that plug-ins should only be used by external people who want to extend the package but not in an open source way - eg the million different repo approach isn’t sustainable (I learned the hard way in bio2bel)

you can always have the testing harness install optional dependencies and we can also make it a bit easier from this class to mock the Gilda grounder. I will provide a small unit test tomorrow

src/oaklib/implementations/gilda.py
cthoyt: On second look, this discussion isn't specific to Gilda so maybe it should be considered in a separate issue

cthoyt:
> lmk your thoughts on the package dependency issue, happy to do this incrementally though

I just made a few commits that show how Gilda can be listed as an optional dependency in the manifest. I added some unit tests that show the code working, and made a tiny update to the CI workflow to install the appropriate optional dependencies. This is the way that many big packages deal with their optional dependencies, so I'd say it's a good first step rather than having to commit or not to spinning these out as plugins.

cmungall (approved):

Issue #193, opened 2022-07-19
wrap fenominal


cmungall:
Phenomenal is a java library for text mining for disease and phenotype concepts.

https://github.com/monarch-initiative/fenominal

This could be wrapped a few ways

1. py4j
2. write a java API and talk to that
3. command line wrapping

PR #199, opened 2022-07-20
SSSOM: Improve matching and implement readily available SSSOM methods to do so.

### LINKED ISSUE #98: Lexmatch only creates skos:closeMatch?
Is there currently any documentation on the lexmatch code?

I noticed that currently everything in `lexmatch` get `skos:closeMatch`:

```
DOID:0050140	acute diarrhea	skos:closeMatch	SYMP:0000675	acute diarrhea	Lexical	oaklib	rdfs:label	rdfs:label	acute diarrhea
```

### LINKED ISSUE #99: lexmatch should create a legal sssom file
Currently I notice these violations:

- License and mapping set ids should be purls (@hrshdhgd we can use what we use as defaults in sssom-py as well)
- no curie map (we should be able to pass in a metadata parameter just as in sssom py)
- Some subject_id/obect_id values are not curies presumable because they are not known w/o a curie map

### LINKED ISSUE #191: Lexmatch revision
This ticket serves as a collection point for the various OAK issues we want to solve in the course of solving https://github.com/monarch-initiative/mondo-ingest/issues/38.

- https://github.com/INCATools/ontology-access-kit/issues/99
- https://github.com/INCATools/ontology-access-kit/issues/98
- https://github.com/INCATools/ontology-access-kit/issues/192

### LINKED ISSUE #192: Improving lexmatch docs
Looking at https://incatools.github.io/ontology-access-kit/cli.html#cmdoption-runoak-lexmatch-L,

- [ ] Add extended documentation of the -L parameter, at least one sentence of what a lexical index file is or where you can learn more.
- [ ] Add a sentence to the docs of the -R parameter, provide a link to where you can learn more about the rule file metadata (maybe examples)
- [ ] --add-labels, explain a tiny bit better what this does, maybe an example.
- [ ] One thing that would be great if we could wrap standard matches like AML and LogMatch in the lexmatch interfaces. Even if it does not turn out feasible right now, we should make sure it is possible in the future, much the same way as we wrap ROBOT for various tasks. I have code to harmonise the outputs of these matches to sssom. These should probably be implemented as a module, and the OAK matcher be conceptualised as a reference implementation. For now, lets just create a page somewhere in the docs "How to extend the `lexmatch` interfaces to implement alternative matchers - this will help us ensure it is at all possibel.
- [ ] From the CLI docs of lexmatch, add see also links to the actual library interfaces

hrshdhgd:
Fix #191 
 - [x] add metadata handling in lexical_index_to_sssom (fixes #99 )
   - [x] License and mapping set ids should be purls 
   - [x] no curie map (we should be able to pass in a metadata parameter just as in sssom py)
   - [ ] Some subject_id/obect_id values are not curies presumable because they are not known w/o a curie map
- [x] documentation on the lexmatch code (fixes #98 )
- [x] More documentation (fixes #192) :
  - [x] Add extended documentation of the -L parameter, at least one sentence of what a lexical index file is or where you can learn more.
  - [x] Add a sentence to the docs of the -R parameter, provide a link to where you can learn more about the rule file metadata (maybe examples)
  - [x] --add-labels, explain a tiny bit better what this does, maybe an example.
  - [ ] ~One thing that would be great if we could wrap standard matches like AML and LogMatch in the lexmatch interfaces. Even if it does not turn out feasible right now, we should make sure it is possible in the future, much the same way as we wrap ROBOT for various tasks. I have code to harmonise the outputs of these matches to sssom. These should probably be implemented as a module, and the OAK matcher be conceptualised as a reference implementation. For now, lets just create a page somewhere in the docs "How to extend the `lexmatch` interfaces to implement alternative matchers - this will help us ensure it is at all possibel.~
  - [x] From the CLI docs of lexmatch, add see also links to the actual library interfaces

### DIFF
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -2256,7 +2256,7 @@ def lexmatch(output, recreate, rules_file, lexical_index_file, add_labels):
 
     See :ref:`.lexical_index_to_sssom`
 
-    If add_labels == True, see :ref:`.add_labels_from_uris`. 
+    If add_labels == True, see :ref:`.add_labels_from_uris`.
     e.g. for URI http://example.org/a/bone_of_foot, label = "bone of foot"
 
     Examples:
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -2237,7 +2237,12 @@ def set_apikey(endpoint, keyval):
     "-L",
     help="path to lexical index. This is recreated each time unless --no-recreate is passed",
 )
-@click.option("--rules-file", "-R", help="path to rules file. Conforms to rules_datamodel.")
+@click.option(
+    "--rules-file",
+    "-R",
+    help="path to rules file. Conforms to rules_datamodel.\
+        e.g. https://github.com/INCATools/ontology-access-kit/blob/main/tests/input/matcher_rules.yaml",
+)
 @click.option(
     "--add-labels/--no-add-labels",
     default=False,
@@ -2274,39 +2279,8 @@ def lexmatch(output, recreate, rules_file, lexical_index_file, add_labels):
 
         runoak  -i foo.obo lexmatch -R match_rules.yaml -L foo.index.yaml -o foo.sssom.tsv
 
-    Example of a LexicalIndex object for the term 'continuant':
-
-    LexicalIndex
-    (
-        groupings=
-        {
-            'continuant': LexicalGrouping
-            (
-                term='continuant',
-                relationships=
-                    [
-                        RelationshipToTerm
-                        (
-                            predicate='rdfs:label',
-                            element='BFO:0000002',
-                            element_term='continuant',
-                            source=None,
-                            pipeline=['default']
-                        )
-                    ]
-            )
-        }
-    )
-
-
-    :param output: Output path.
-    :param recreate: If true and lexical index is specified, always recreate, otherwise load from index.
-    :param rules_file: Rules of matching. This is a YAML file
-        e.g. https://github.com/INCATools/ontology-access-kit/blob/main/tests/input/matcher_rules.yaml
-    :param lexical_index_file: Path to lexical index. This is recreated each time unless --no-recreate is passed.
-    :param add_labels: Populate empty labels with URI fragments or CURIE local IDs,
-        for ontologies that use semantic IDs.
-    :raises NotImplementedError: If implementation is anything other than BasicOntologyInterface.
+    For more on LexicalIndex,
+    see https://incatools.github.io/ontology-access-kit/datamodels/lexical-index/index.html.
     """
     impl = settings.impl
     if rules_file:
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -2261,9 +2261,6 @@ def lexmatch(output, recreate, rules_file, lexical_index_file, add_labels):
 
     See :ref:`.lexical_index_to_sssom`
 
-    If add_labels == True, see :ref:`.add_labels_from_uris`.
-    e.g. for URI http://example.org/a/bone_of_foot, label = "bone of foot"
-
     Examples:
 
         runoak -i foo.obo lexmatch -o foo.sssom.tsv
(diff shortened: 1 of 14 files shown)


hrshdhgd:
@cmungall , @matentzn  I think I've addressed the ones check marked. Amongst the other 2 :

 - Some subject_id/obect_id values are not curies presumable because they are not known w/o a curie map
   - For something like `http://example.org/a/bone_of_foot` , there is no `curie_map`(`prefix_map`) in defaults either. Example is [this owl file](https://github.com/INCATools/ontology-access-kit/blob/main/tests/input/matcher-test.owl).
 - The "AML and LogMatch", don't quite know what do I do with that.

src/oaklib/cli.py
```diff
@@ -2051,7 +2051,7 @@ def add_taxon_constraints(constraints, evolution_file, predicates: List, output)
         impl.enable_transitive_query_cache()
         for st in sts:
             try:
-                st = test_candidate_taxon_constraint(impl, st, predicates=actual_predicates)
+                st = eval_candidate_taxon_constraint(impl, st, predicates=actual_predicates)
```
cmungall, line 2054: this is leaking over from a different PR?

hrshdhgd: No I manually made this change when I discovered it working with Justin and Harry.

src/oaklib/cli.py
```diff
@@ -2271,6 +2273,40 @@ def lexmatch(output, recreate, rules_file, lexical_index_file, add_labels):
     Using custom rules:
 
         runoak  -i foo.obo lexmatch -R match_rules.yaml -L foo.index.yaml -o foo.sssom.tsv
+
+    Example of a LexicalIndex object for the term 'continuant':
+
+    LexicalIndex
```
cmungall, line 2279: I think this is way too much for the CLI docs. This isn't even a real use case for doing things with this, any more than there is for introspecting a blast index. Most of the time the user doesn't need to know this exists. On some occasions for large ontologies it's use to persist this to do repeated runs without re-indexing, but you don't even need to look at it.

The right place to put this is in the **datamodel documentation**. All objects in OAK conform to a datamodel. This one is documented here: https://incatools.github.io/ontology-access-kit/datamodels/lexical-index/index.html

The CLI docs should just point to this. Adding examples to this would be great, but do it in the yaml!

cc @matentzn

src/oaklib/cli.py
```diff
@@ -2271,6 +2273,40 @@ def lexmatch(output, recreate, rules_file, lexical_index_file, add_labels):
...
+                            source=None,
+                            pipeline=['default']
+                        )
+                    ]
+            )
+        }
+    )
+
+
+    :param output: Output path.
```
cmungall, line 2302: This is a CLI function

src/oaklib/cli.py
```diff
@@ -2252,11 +2252,13 @@ def set_apikey(endpoint, keyval):
 )
 @output_option
 def lexmatch(output, recreate, rules_file, lexical_index_file, add_labels):
-    """
-    Generates lexical index and mappings
+    """Generates lexical index and mappings.
 
     See :ref:`.lexical_index_to_sssom`
 
+    If add_labels == True, see :ref:`.add_labels_from_uris`.
```
cmungall, line 2259: this documentation aimed at python developers, but it's a CLI docstring

cmungall:
> @cmungall , @matentzn I think I've addressed the ones check marked. Amongst the other 2 :
> 
> * Some subject_id/obect_id values are not curies presumable because they are not known w/o a curie map
>   
>   * For something like `http://example.org/a/bone_of_foot` , there is no `curie_map`(`prefix_map`) in defaults either. Example is [this owl file](https://github.com/INCATools/ontology-access-kit/blob/main/tests/input/matcher-test.owl).

We want to make everything easy for people to do matching with arbitrary ontologies, not just OBO. For now we will amend the test to inject a prefix

> * The "AML and LogMatch", don't quite know what do I do with that.

Not in this PR. The request here is just to have plugins for common mapping tools just as we are planning plugins for annotators e.g. #193

cmungall (approved):

PR #202, opened 2022-07-20
Renaming test_candidate_taxon_constraint method

justaddcoffee:


justaddcoffee:
@hrshdhgd @caufieldjh

hrshdhgd:
running `tox` and then committing the files should fix the error.

cmungall (approved):

cmungall:
Thanks for this, not only was this name messing with pytest, it was following poor naming conventions. Your change improves it muchly thanks!

Issue #173, opened 2022-07-08
Reduce `print()` calls in tests


cthoyt:
Right now it's pretty difficult to run the tests since they create such a huge amount of output. Ideally, the print statements in tests can be switched to using the logger module, then the test harness only shows the results from failed tests.

pkalita-lbl:
I think this can be closed now that #210 is merged. @cthoyt had some good follow-on tasks in the comments of that PR. I can make issues for those.

PR #142, opened 2022-06-24
Demonstrate using PyStow for API configuration

cthoyt:
- [x] Replace builtin code for configuration access and writing
- [x] Write documentation on how this works

PR #204, opened 2022-07-21
Use ontoportal-client classes in BioPortal and AgroPortal implementations

### LINKED ISSUE #168: Externalize OntoPortal client functions (e.g., for BioPortal, AgroPortal, EcoPortal)
I have previous written a harness for accessing the API for OntoPortal instances in the Bioregistry project. I spun that code out into a clean project at https://github.com/cthoyt/ontoportal-client. I'd be happy to move that to a different namespace if other people want to contribute to it.

pkalita-lbl:
These changes introduce an `OntoPortalInterface` base class. It expects subclasses to populate the `OntoPortalClientClass` and `api_key_name` class variables. `OntoPortalInterface` then handles all of the implementation details of constructing the `OntoPortalClient` instance and using it to make requests.

Fixes #168


cmungall (approved):
I wasn't thinking of OntoPortal as an interface, but rather as a parent class for some implementations.

A few other things to bear in mind:

- ontoportal-client is still young, I figure the methods it exposes may vary over time, which will mean a bit more coordinating and perhaps some version pinning
- we don't have a great testing strategy for bioportal and friends yet. I run the tests locally using my API key, but gh actions will skip them (I think). We need to set up Mock unit tests, and have separate integration tests. This will all come in time but just be aware that there is more potential to temporarily break things here!

pkalita-lbl:
I agree there is some potential for confusion here about `OntoPortalInterface`; it's not exactly an interface in the same way the other classes in the `interfaces` package are. I considered putting it in the `implementations` package, but it's also not quite an implementation that can can be used on its own. It's a bit neither-here-nor-there, but thinking about it now it probably belongs more in `implementations`. I'll make that change before merging. 

Definitely on the same page as you about testing. I also run them locally using my API key, and you're right that they are skipped when run in the GitHub workflow. Mocking can be a headache but it's better than nothing. I can open an issue about that.

cthoyt:
The ontoportal client has some pretty slick testing set up, you can see how the combination of api keys set as GitHub actions secrets and the unified lookup with pystow simplifies it pretty nicely

pkalita-lbl:
That's also a good idea @cthoyt. I'll take a closer look at that when I work on #205. 

I've updated this branch to replace `oaklib.interfaces.ontoportal_interface.OntoPortalInterface` with `oaklib.implementations.ontoportal.ontoportal_implementation_base.OntoPortalImplementationBase`. I think that makes it more clear what the purpose of that class is.

Maybe long term we want to put the BioPortal, AgroPortal, etc implementations into the `oaklib.implementations.ontoportal` package as well?

cthoyt:
@pkalita-lbl just pushed a new ontoportal-client release (https://github.com/cthoyt/ontoportal-client/releases/tag/v0.0.3), that also does more of the things that you have in your get_response and get_json wrappers to reduce more code

src/oaklib/implementations/bioportal/agroportal_implementation.py
```diff
@@ -1,23 +1,18 @@
...
     """
 
-    @property
-    def _base_url(self) -> str:
-        return "http://data.agroportal.lirmm.fr/"
-
-    def load_bioportal_api_key(self, path: str = None) -> None:
-        self.bioportal_api_key = "1de0a270-29c5-4dda-b043-7c3580628cd5"
+    OntoPortalClientClass = AgroPortalClient
+    api_key_name = "agroportal"
```
cthoyt, line 18: this should be inferrable from the client class now with ontoportal-client v0.0.3

src/oaklib/implementations/bioportal/agroportal_implementation.py
```diff
@@ -1,23 +1,18 @@
...
 
     """
 
-    @property
-    def _base_url(self) -> str:
-        return "http://data.agroportal.lirmm.fr/"
-
-    def load_bioportal_api_key(self, path: str = None) -> None:
-        self.bioportal_api_key = "1de0a270-29c5-4dda-b043-7c3580628cd5"
+    OntoPortalClientClass = AgroPortalClient
```
cthoyt, line 17: would suggest following pep8 naming conventions and call this `ontoportal_client_class` instead

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
```diff
@@ -0,0 +1,271 @@
...
+        response = self.client.get_response(path, *args, **kwargs)
+        if raise_for_status:
+            response.raise_for_status()
+        return response
+
+    def _get_json(self, *args, **kwargs):
+        response = self._get_response(*args, **kwargs)
+        return response.json()
+
+    def ontologies(self) -> Iterable[CURIE]:
```
cthoyt, line 74: the type annotation here for the result is misleading. the acronyms are prefixes, not CURIEs

cmungall: good point, I think this is something that may need a deeper fix as part of a separate PR through

src/oaklib/implementations/bioportal/agroportal_implementation.py
pkalita-lbl: JavaScript habits are hard to shake 😁 Will fix.

cthoyt: there are code quality checks to ensure we don't do stuff like this by accident, some of them are just taking longer to put in place in earnest in this project because it already has so much technical debt

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
```diff
@@ -50,22 +49,18 @@ def __post_init__(self):
         if self.focus_ontology is None:
             if self.resource:
                 self.focus_ontology = self.resource.slug
-        if not self.OntoPortalClientClass:
-            raise NotImplementedError("OntoPortalClientClass not specified")
-        api_key = get_apikey_value(self.api_key_name)
-        self.client = self.OntoPortalClientClass(api_key=api_key)
+        if not self.ontoportal_client_class:
+            raise NotImplementedError("ontoportal_client_class not specified")
+        api_key = get_apikey_value(self.ontoportal_client_class.name)
```
cthoyt, line 54: the API key getting is handled by the class itself, so you can skip this completely from the OAK side

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
```diff
@@ -50,22 +49,18 @@ def __post_init__(self):
...
+        self.client = self.ontoportal_client_class(api_key=api_key)
 
     def prefix_map(self) -> PREFIX_MAP:
         # TODO
         return {}
 
-    def _get_response(self, path: str, *args, raise_for_status=True, **kwargs):
+    def _get_response(self, *args, **kwargs):
         check_limit()
```
cthoyt, line 62: do you know if the rate limit is the same for all instances? we could also push the limit checking upstream

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
pkalita-lbl: I think I'm with Chris on this one. It's not good, but it's also not an isolated thing and would probably be handled better in a separate PR.

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
pkalita-lbl: Yeah I saw that ontoportal_client manages API keys as well but as I understand it that would require the user to create either an environment variable or a file under `~/.config`. The OAK command line interface has a `set-apikey` subcommand that I think we need to continue to support. So I thought it was better to continue to let OAK manage its own key storage.

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
pkalita-lbl: It's basically a global rate limiter: https://github.com/INCATools/ontology-access-kit/blob/main/src/oaklib/utilities/rate_limiter.py which seems perhaps too conservative, but there may be some good reasons for it that I'm not aware of.

cthoyt: By instances, I mean bioportal, agoportal, etc. I know BioPortal says 15 requests per second on https://www.bioontology.org/wiki/Annotator_Optimizing_and_Troublehooting, but finding this kind of specific documentation for all portals is annoying

pkalita-lbl: Ah I see, no I don't know off hand if all ontoportal instances have the same rate limits or if it's configurable.

cmungall:
Yes, consider this a first pass. Globally it's conservative as some of the
individual queries launched at the triplestore endpoints can be
mega-queries, but it could be tuned for each category of endpoint, then by
endpoint specifics

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
cthoyt: In https://github.com/INCATools/ontology-access-kit/pull/142 I made it much more integrated with PyStow, since it's a pain to have to use several competing systems when there's already something ready to go out of the box. If someone wants to put effort into unmucking the testing (#173), then it will be possible to finish that PR, since right now the logging is too much of a mess to even figure it out. If I had to guess though, it might be because PRs from forks don't get access to the secrets in the GHA workflows

cmungall:
I think using pystow universally for api keys is good. Up to you both how
best to go about it ie merge other pr, redo work on this one

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
pkalita-lbl: I can talk to Sujay about #173 since he currently is assigned and see if has any progress on it. If not maybe I can take it over from him. Also after I commented on #142 Chris gave me admin rights on this repo, so I'd be happy to work with you on that one. But as far as these changes go are you okay with decoupling them from the PyStow work? I'm not sure how much effort that will be, but does it make sense to just get this in now?

cthoyt: All of the code is simple, the only real effort is coordination :p

cthoyt: I can clean this up later, it’s not a blocker for your PR

pkalita-lbl: Definitely appreciate all your feedback. I'm going to merge this now, but please get in touch with me next week (obo-community slack maybe?) and we can try to coordinate on getting #142 going.

PR #210, opened 2022-07-25
Reduce output noise in tests

pkalita-lbl:
These changes reduce test output noise by replacing calls to `print` with `logging.info` and using the special value `'-'` as the default for the `--output` option. Replacing `sys.stdout` with `'-'` doesn't have any user-facing consequence, but it ensures that output gets properly captured when running tests via `CliRunner`.


pkalita-lbl:
@cthoyt I took a crack at fixing #173 here. The output from running tests is now greatly reduced. There are still some warnings generated by OAK itself. Those could potentially be suppressed by setting the logging level in `tests/__init__.py`. There are also some warnings coming from outside OAK that I'm not totally sure about. Thoughts?

cthoyt:
Cool, this is a great step forward so we might as well merge it ASAP rather than getting a perfect solution.

pkalita-lbl:
Okay great. If you don't mind hitting the approve button I will merge it in.

cthoyt (approved):
Fine for now - a few potential improvements:
1. use a specific logger instead of the top-level logging.info function
2. delete dead code, and improve tests where the only thing that happened was logging
3. use pytest instead of unittest, I think it handles logging better

PR #215, opened 2022-08-02
Get 'scope' information from the command.

hrshdhgd:
Previous iteration #209 assumed all synonyms to be `RELATED`. Now this can be derived from the command using the following code:
```
scope = str(patch.qualifier.value).upper() if patch.qualifier else "RELATED"
t.add_synonym(description=patch.new_value, scope=scope)
```

PR #209, opened 2022-07-22
New synonym request handling  for `ProntoImplementation`

hrshdhgd:
fixes #208

PR #217, opened 2022-08-03
Revert "Get 'scope' information from the command."

hrshdhgd:
Reverts INCATools/ontology-access-kit#215

Previous iteration #209 assumed all synonyms to be `RELATED`. Now this can be derived from the command using the following code:
```
scope = str(patch.qualifier.value).upper() if patch.qualifier else "RELATED"
t.add_synonym(description=patch.new_value, scope=scope)
```

### DIFF
--- a/src/oaklib/implementations/pronto/pronto_implementation.py
+++ b/src/oaklib/implementations/pronto/pronto_implementation.py
@@ -521,7 +521,7 @@ class ProntoImplementation(
                     syn.description = patch.new_value
         elif isinstance(patch, kgcl.NewSynonym):
             t = self._entity(patch.about_node, strict=True)
-            # Get scope from patch.qualifier 
+            # Get scope from patch.qualifier
             # rather than forcing all synonyms to be related.
             scope = str(patch.qualifier.value).upper() if patch.qualifier else "RELATED"
             t.add_synonym(description=patch.new_value, scope=scope)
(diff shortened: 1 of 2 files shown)


src/oaklib/implementations/pronto/pronto_implementation.py
```diff
@@ -520,6 +520,9 @@ def _clean(v: str) -> str:
                 if syn.description == patch.old_value:
                     syn.description = patch.new_value
         elif isinstance(patch, kgcl.NewSynonym):
+            t = self._entity(patch.about_node, strict=True)
```
hrshdhgd, line 523: This line was already present before. For some reason the reverting of #215 messed things up. The only reason I have this PR is to bring the change in lines 526 and 527 to Chris' attention.

src/oaklib/implementations/pronto/pronto_implementation.py
```diff
@@ -520,6 +520,9 @@ def _clean(v: str) -> str:
...
+            # Get scope from patch.qualifier 
+            # rather than forcing all synonyms to be related.
             scope = str(patch.qualifier.value).upper() if patch.qualifier else "RELATED"
             t.add_synonym(description=patch.new_value, scope=scope)
```
hrshdhgd, line 528: The line 523 was already present before. For some reason the reverting of https://github.com/INCATools/ontology-access-kit/pull/215 messed things up. The only reason I have this PR is to bring the change in lines 526 and 527 to Chris' attention.

PR #224, opened 2022-08-12
Externalize CURIE/IRI conversion code

### LINKED ISSUE #214: `curie_from_iri()` issues
## Overview
I know that we are moving to OAK, but I've been having a lot of trouble getting my OAK use cases to work so far, and [@matentzn  recommended](https://github.com/monarch-initiative/mondo-ingest/pull/35#discussion_r918476599) that I tried the `bioregistry` function first. However I had problems w/ that, so I opened up 2 issues: (https://github.com/biopragmatics/bioregistry/issues/479, https://github.com/biopragmatics/bioregistry/issues/480)

I then went on to use OAK, but had some issues here as well.

### 1. Fails when angle brackets present in URI
I don't have a code snippet for this, sorry. But I did not get the correct result.

### 2. Duplicated code
This may not be comprehensive, but I see the same block of code for this in `BasicOntologyInterface`, `FunOwlImplementation`, and `AbstractSparqlImplementation`.

### 3. No static function
It shouldn't be necessary to load up an ontology instantiate an implementation / interface object. If I have a single URI string, and a `prefix_map`, I should be able to use OAK in a simple way by using 1 line of code to import the `uri_to_curie` and a second line to call the function.

There should probably be some area in OAK to store such static functions. Then, the classes can wrap around that; something like:
```py
from oaklib.utils import uri_to_curie

def uri_to_curie(self, uri: URI, strict=True) -> Optional[CURIE]:
    pm = self.get_prefix_map()
    return uri_to_curie(uri, pm)
```

Better yet, put it in `OntologyInterface` (assuming all the implementation / interfaces in inherit from that) and even that wrapper method needs only be declared once.

cthoyt:
Closes #214

This PR replaces the default implementations for `curie_to_uri()` and `uri_to_curie()` using the prefixes returned by `get_prefixes()` in combination with the [`curies`](https://github.com/cthoyt/curies) package.

CC @joeflack4

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ kgcl-rdflib = "^0.2.0"
 pystow = "^0.4.4"
 class-resolver = "^0.3.10"
 ontoportal-client = "0.0.3"
+curies = "^0.1.5"
 
 [tool.poetry.dev-dependencies]
 pytest = "^5.2"
--- a/src/oaklib/interfaces/basic_ontology_interface.py
+++ b/src/oaklib/interfaces/basic_ontology_interface.py
@@ -3,16 +3,10 @@ from abc import ABC
 from dataclasses import field
 from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple
 
+import curies
 from deprecation import deprecated
 
-from oaklib.datamodels.vocabulary import (
-    BIOPORTAL_PURL,
-    IS_A,
-    OBO_PURL,
-    OWL_CLASS,
-    OWL_NOTHING,
-    OWL_THING,
-)
+from oaklib.datamodels.vocabulary import IS_A, OWL_CLASS, OWL_NOTHING, OWL_THING
 from oaklib.interfaces.ontology_interface import OntologyInterface
 from oaklib.types import CURIE, PRED_CURIE, SUBSET_CURIE, URI
 from oaklib.utilities.basic_utils import get_curie_prefix
@@ -80,6 +74,7 @@ class BasicOntologyInterface(OntologyInterface, ABC):
 
     strict: bool = False
     autosave: bool = field(default_factory=lambda: True)
+    _converter: curies.Converter = field(init=False, repr=False)
 
     def prefix_map(self) -> PREFIX_MAP:
         """
@@ -93,7 +88,17 @@ class BasicOntologyInterface(OntologyInterface, ABC):
     def get_prefix_map(self) -> PREFIX_MAP:
         return self.prefix_map()
 
-    def curie_to_uri(self, curie: CURIE, strict=False) -> Optional[URI]:
+    @property
+    def converter(self) -> curies.Converter:
+        """Get a converter for this ontology interface's prefix map.
+
+        :return: A converter
+        """
+        if self._converter is None:
+            self._converter = curies.Converter.from_prefix_map(self.prefix_map())
+        return self._converter
+
+    def curie_to_uri(self, curie: CURIE, strict: bool = False) -> Optional[URI]:
         """
         Expands a CURIE to a URI
 
@@ -101,24 +106,12 @@ class BasicOntologyInterface(OntologyInterface, ABC):
         :param strict: (Default is False) if True, exceptions will be raised if curie cannot be expanded
         :return:
         """
-        if curie.startswith("http"):
-            return curie
-        pm = self.prefix_map()
-        parts = curie.split(":")
-        if len(parts) == 2:
-            pfx, local_id = parts
-        else:
-            if strict:
-                raise ValueError(f"Bad CURIE: {curie} parts: {parts}")
-            else:
-                return curie
-        if pfx in pm:
-            return f"{pm[pfx]}{local_id}"
-        else:
-            # TODO: not hardcode
-            return f"{OBO_PURL}{pfx}_{local_id}"
+        rv = self.converter.expand(curie)
+        if rv is None and strict:
+            raise ValueError
+        return rv
 
-    def uri_to_curie(self, uri: URI, strict=True) -> Optional[CURIE]:
+    def uri_to_curie(self, uri: URI, strict: bool = True) -> Optional[CURIE]:
         """
         Contracts a URI to a CURIE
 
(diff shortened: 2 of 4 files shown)


cthoyt:
After working through this, I've realized that the URI contraction and CURIE expansion functionalities were only working because of hacking in hard-coded processing of BioPortal and OBO PURL processing, not because the ontology interfaces were built to generate meaningful prefix maps.

cthoyt:
After doing some archaeology, I found tons of hacked up different implementations of URI/CURIE conversions. Here's how I went about addressing this:

1. Add a principled default prefix map to the base ontology interface class that contains the OBO PURL context, the semantic web context, and a few extras (e.g., EFO, SCTID)
2. Remove all of the implementations in base classes that don't do anything (e.g., return an empty dict, just raise an error, etc.). To be honest, it's making me seriously question some of the choices about mixins and multiple inheritance that don't actually use the `abc.ABC` machinery since a lot of these seemed like "i'll do it later" type issues
3. Replaced many of the URI/CURIE conversion functions which all basically implemented the same thing, +/- the randomness induced by coffee on each day each was implemented separately
4. Some of the changes in 3) had effects on the FunOWL and Pronto that I could not trace back, due to the deep nested nature of how these implementations work and minimal testing. Unfortunately the only solution I could see to updating their processing was to better handle errors rather than improve it to handle all cases in the test code

src/oaklib/implementations/sparql/abstract_sparql_implementation.py
```diff
@@ -119,17 +119,13 @@ def _default_url(self) -> str:
...
-        # return {'rdfs': str(RDFS)}
-        return DEFAULT_PREFIX_MAP
-
     # def store(self, resource: OntologyResource) -> None:
     #    SparqlBasicImpl.dump(self.engine, resource)
 
     def curie_to_uri(self, curie: CURIE, strict: bool = False) -> URI:
         if curie.startswith("http"):
             return curie
+        # FIXME replace with super() call
```
cthoyt, line 128: Punted for next time

cmungall:
See

- https://github.com/cthoyt/curies/issues/16
- https://github.com/cthoyt/curies/issues/17

cthoyt:
I wasn't able to debug the reasons why the Bioregistry's copy of the OBO context wasn't working (probably some mismatch of assumptions to test data) so I replaced it with an internal copy of the OBO prefix map, and now things seem to look alright.

cthoyt:
also, I checked in on the issues posted on the `curies` repository - I don't think any of them are reasons to block this PR. Obviously this isn't a perfect solution, but what existed before was problematic and incomplete, so at least it's a major step forwards.

cmungall:
I'll have a variant of this solution shortly...

joeflack4:
This may also help address: #236

cmungall:
This is on the right lines, some changes I would like to make:

1. add a dependency to https://github.com/linkml/prefixmaps
2. don't cache obo or other prefix maps within this repo - use prefixmaps
3. keep everything use curie.Converter, but don't do composition within this, use prefixmaps for composition

I am not sure if the best approach is

1. keep working on this PR
2. make a new PR
3. branch this PR

cthoyt:
i can take care of those things in this PR

cthoyt:
> This is on the right lines, some changes I would like to make:
> 
>     1. add a dependency to https://github.com/linkml/prefixmaps
> 
>     2. don't cache obo or other prefix maps within this repo - use prefixmaps
> 
>     3. keep everything use curie.Converter, but don't do composition within this, use prefixmaps for composition

All of these things are now done

pyproject.toml
```diff
@@ -27,6 +27,9 @@ kgcl-rdflib = "^0.3.0"
 pystow = "^0.4.4"
 class-resolver = "^0.3.10"
 ontoportal-client = "0.0.3"
+curies = "^0.1.5"
```
hrshdhgd, line 30: [prefixmaps already has a curies dependency](https://github.com/linkml/prefixmaps/blob/bdd770a5f65c6fc16a229d9632771db659b30a54/pyproject.toml#L16). Isn't this redundant?

cthoyt: no it should definitely be explicit since this package invokes `curies` directly

cmungall: agreed... and I don't think prefixmaps needs that dependency. It's literally just a dumb named prefix->namespace mapping cache

src/oaklib/implementations/sparql/abstract_sparql_implementation.py
cmungall: np, I know how to simplify this

src/oaklib/interfaces/basic_ontology_interface.py
```diff
@@ -18,13 +20,29 @@
...
 NC_NAME = str
-PREFIX_MAP = Dict[NC_NAME, URI]
+PREFIX_MAP = Mapping[NC_NAME, URI]
 RELATIONSHIP_MAP = Dict[PRED_CURIE, List[CURIE]]
 ALIAS_MAP = Dict[PRED_CURIE, List[str]]
 METADATA_MAP = Dict[PRED_CURIE, List[str]]
 # ANNOTATED_METADATA_MAP = Dict[PRED_CURIE, List[Tuple[str, METADATA_MAP]]]
 RELATIONSHIP = Tuple[CURIE, PRED_CURIE, CURIE]
 
+MISSING_PREFIX_MAP = dict(
```
cmungall, line 30: I'll take care of this on a future round

src/oaklib/interfaces/basic_ontology_interface.py
```diff
@@ -80,45 +98,54 @@ class BasicOntologyInterface(OntologyInterface, ABC):
...
-            if strict:
-                raise ValueError(f"Bad CURIE: {curie} parts: {parts}")
-            else:
-                return curie
-        if pfx in pm:
-            return f"{pm[pfx]}{local_id}"
-        else:
-            # TODO: not hardcode
-            return f"{OBO_PURL}{pfx}_{local_id}"
+        rv = self.converter.expand(curie)
```
cmungall, line 136: minor: maybe this logic could be moved up to converter

cmungall (approved):
This is fantastic, thanks so much @cthoyt!!

PR #263, opened 2022-09-08
Add a check of ENV variables to the apikey_manager.get_apikey_value method  fixes #262

### LINKED ISSUE #262: add a check for API keys in the environment in get_apikey_value method
currently, this checks for a specifically formatted text file with the key; I'd like to store the key in my env and have this method find it. 
I'll put in a PR for review.

sierra-moxon:
Using oak in schema_automator to access bioportal with BIOPORTAL_API_KEY set in the ENV.


hrshdhgd (approved):
Looks good to me!

PR #264, opened 2022-09-08
Change collections import to abide by new package design

hrshdhgd:
~from collections import Iterator, defaultdict~
```
from collections import defaultdict
from collections.abc import Iterator
```

### DIFF
--- a/src/oaklib/utilities/mapping/cross_ontology_diffs.py
+++ b/src/oaklib/utilities/mapping/cross_ontology_diffs.py
@@ -19,7 +19,7 @@ from oaklib.utilities.graph.networkx_bridge import mappings_to_graph
 from oaklib.utilities.label_utilities import add_labels_to_object
 
 try:
-    from collections import Iterator, defaultdict
+    from collections import Iterator, defaultdict  # noqa:F811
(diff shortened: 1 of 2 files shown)


src/oaklib/utilities/mapping/cross_ontology_diffs.py
```diff
@@ -18,6 +17,12 @@
 from oaklib.utilities.graph.networkx_bridge import mappings_to_graph
 from oaklib.utilities.label_utilities import add_labels_to_object
 
+try:
```
cmungall, line 20: super minor but IMO it's cleaner not to wrap the defaultdict import as it's nonproblematic

it's just Iterator
https://stackoverflow.com/questions/72032032/importerror-cannot-import-name-iterable-from-collections-in-python

cmungall: in fact do do you need to wrap at all?

Doesn't this work in 3.9?
from collections.abc import Iterator

hrshdhgd: You're right. It seems to work. Changing code

cmungall (approved):

PR #270, opened 2022-09-09
Externalize OLS implementation

cthoyt:
This PR switches out interaction with the EBI's ontology lookup service using the [`ols-client`](https://github.com/cthoyt/ols-client/) package, similarly to how we externalized some functionality for OntoPortal instances in #204.

Further, this enables interaction with alternative OLS instances like the TIB Hannover one. Probably of interest to @StroemPhi.


StroemPhi:
Oooh cool! Thank you @cthoyt, am definitely interested!
Although I must admit, I didn't have the time yet to actually test OAK. I only read the paper draft and thus only have a sense of how useful it can be. I did bring this up to Oliver and our devs, but will propably need to carve out some more prio points for this somehow. ;)

cmungall:
Thanks @cthoyt this is awesome! I have no idea why that 3.10 test is failing

@StroemPhi if you can install a python module you can be doing awesome things on the command line in seconds, ping me on slack with what kind of thing you’d find most useful

cmungall:
I think rebase should fix (gets rid of the problematic switch on the nx code) but odd how it got into this state in the first place

cmungall (approved):

PR #273, opened 2022-09-12
Combine coverage analysis

cthoyt:
The tests were running twice, which is wasteful of compute resources and slow. This PR combines them together. It also switches from unittest to pytest, which makes it much easier to look through the logs


cmungall (approved):

cmungall:
What about the bioportal secrets change?

cmungall:
d'oh, I didn't realize the key was duplicated before, I just looked at the diff and it looked like it was being removed altogether

PR #282, opened 2022-09-16
Delay construction of ClassResolver instance

### LINKED ISSUE #280: Plugin discovery is sensitive to circular imports
Currently the oakx-grape plugin is registered as expected. However, if [this line ](https://github.com/INCATools/oakx-grape/blob/9e92c86891f439d1c46560ccddf7206a85f58374/src/oakx_grape/grape_implementation.py#L8) is altered to include importing `get_implementation_from_shorthand` attempting to use the `grape:` input scheme fails with an error saying that the scheme is not recognized. 

The reason this is happening is not that the oakx-grape plugin is doing anything wrong per se. Rather, `implementation_resolver` is being [defined](https://github.com/INCATools/ontology-access-kit/blob/4d1ab9e0bbe7668161d927c0d3fd76f082731f6c/src/oaklib/implementations/__init__.py#L62) at the time that `oaklib.implementations` is imported. And when `implementation_resolver.register_entrypoint` attempts to load the plugin classes it fails with a partially initialized module (circular imports) error. class-resolver [swallows that error](https://github.com/cthoyt/class-resolver/blob/cffb9cdb7f3c84d0bd3ce6551af63f7345fed8c4/src/class_resolver/base.py#L253), and the end result is that the plugin's scheme is not registered.

pkalita-lbl:
These changes move the construction of the `ClassResolver` instance for implementation classes into a function. This allows the loading of entry point classes to be delayed until one is actually needed. And that should make it a lot harder to get into the circular import / partially initialized module trap. The function is memoized so only one `ClassResolver` instance is ever created.

Fixes #280

### DIFF
--- a/src/oaklib/implementations/__init__.py
+++ b/src/oaklib/implementations/__init__.py
@@ -1,4 +1,5 @@
 from functools import cache
+
 from class_resolver import ClassResolver
 
 from oaklib.implementations.funowl.funowl_implementation import FunOwlImplementation
@@ -59,6 +60,7 @@ __all__ = [
     "FunOwlImplementation",
     "GildaImplementation",
 ]
+
 
 @cache
 def get_implementation_resolver() -> ClassResolver[OntologyInterface]:
--- a/src/oaklib/implementations/__init__.py
+++ b/src/oaklib/implementations/__init__.py
@@ -1,3 +1,4 @@
+from functools import cache
 from class_resolver import ClassResolver
 
 from oaklib.implementations.funowl.funowl_implementation import FunOwlImplementation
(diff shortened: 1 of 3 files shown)


cthoyt:
I would highly suggest spending your effort to untangle the circular dependencies instead

cthoyt:
Alternative I'd accept a PR to improve the ClassResolver to make plugin loading lazy

pkalita-lbl:
> I would highly suggest spending your effort to untangle the circular dependencies instead

I made an attempt at that, but I convinced myself that there is something fundamentally circular about this situation. OAK plugins will generally import something from OAK and OAK needs to import the entry point classes from the plugins. If it's all happening at import-time, it's going to be dicey. Maybe it would be possible to shuffle around the `ClassResolver` to another module or something? But I feel like that leaves things a bit fragile. And, to me, delaying the instantiation is a benefit unto itself.

> Alternative I'd accept a PR to improve the ClassResolver to make plugin loading lazy

I can try to take a look at that next week.

cmungall (approved):

cmungall:
So I think we are good to go with this PR, and we can also do a PR further upstream too

PR #287, opened 2022-09-21
Bugfix: basic_ontology_interface

joeflack4:
Updates
```
Bugfix: basic_ontology_interface
- Bugfix: hierararchical_parents(): If no parents found, now returns an empty list rather than resulting in `KeyError`.
```


joeflack4:
Thanks!

PR #303, opened 2022-10-05
Fix py version window & provide minimum scipy version

### LINKED ISSUE #300: Restore gilda test
This was skipped in order to resolve #299, as the gilda test had stopped running on 3.10

hrshdhgd:
This fixes #300 

Problem: By default an older SciPy version (dependency of a dependency) is installed (v1.6.1) which does not play well with python v3.10. So I add it as a direct dependency and anchor it to `>=1.9.1`. On doing that, there is a complain raised stating the project python version compatibility is from 3.9 to 4 but with the scipy version installed it is 3.9 to 3.12. So I set python version window between 3.9 and 3.12 and now everything seems to be okay.

This window will need to be revisited while supporting future version of python.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -18,7 +18,7 @@ test = ["pytest", "pytest-cov"]
 
 [[package]]
 name = "aiohttp"
-version = "3.8.1"
+version = "3.8.3"
 description = "Async http client/server framework (asyncio)"
 category = "main"
 optional = false
@@ -201,7 +201,7 @@ lxml = ["lxml"]
 
 [[package]]
 name = "bioregistry"
-version = "0.5.94"
+version = "0.5.119"
 description = "Integrated registry of biological databases and nomenclatures"
 category = "main"
 optional = false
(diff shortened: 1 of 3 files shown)


cmungall (changes requested):
I don't want to make scipy a dependency of OAK, it's a heavyweight dependency

Note: it's not currently a main dependency, currently gilda is an extra, and scipy is a dependency on that extra

Can we move it to extras?

Alternatively this may be simpler if we make gilda a plugin

hrshdhgd:
Done.

cmungall (approved):
I'm missing something

Gilda already depends on scipy, no?

The original point was to ensure a minimum version for scipy?

But it passes...

hrshdhgd:
Yes , by mentioning `scipy` before `gilda` , it installs the latest package rather than depending on `gilda` deciding it. For some reason it installed v1.6.1 inspire of v1.91 being the latest version.

PR #317, opened 2022-10-18
Added `qualifier` attr to Synonymizer

hrshdhgd:
Added `qualifier` attribute to Synonymizer to indicate if the new synonymized label is an `exact`, `broad`, `narrow` or `related` synonym.

### DIFF
--- a/src/oaklib/utilities/lexical/lexical_indexer.py
+++ b/src/oaklib/utilities/lexical/lexical_indexer.py
@@ -46,6 +46,7 @@ from oaklib.types import PRED_CURIE
 from oaklib.utilities.basic_utils import pairs_as_dict
 
 LEXICAL_INDEX_FORMATS = ["yaml", "json"]
+DEFAULT_QUALIFIER = "exact"
 
 
 def add_labels_from_uris(oi: BasicOntologyInterface):
@@ -124,7 +125,7 @@ def create_lexical_index(
                     term2 = term
                     for tr in pipeline.transformations:
                         if tr.type.code == TransformationType.Synonymization:
-                            synonymized, term2 = apply_transformation(term2, tr)
+                            synonymized, term2, _ = apply_transformation(term2, tr)
                         else:
                             term2 = apply_transformation(term2, tr)
 
@@ -406,14 +407,18 @@ def apply_transformation(term: str, transformation: LexicalTransformation) -> st
         )
 
 
-def apply_synonymizer(term: str, rules: List[Synonymizer]) -> str:
+def apply_synonymizer(term: str, rules: List[Synonymizer]) -> Tuple:
     tmp_term = term
     for rule in rules:
         term = re.sub(eval(rule.match), rule.replacement, term)
+        if rule.qualifier:
+            qualifier = rule.qualifier
+        else:
+            qualifier = DEFAULT_QUALIFIER
     if tmp_term == term:
-        return False, term.rstrip()
+        return False, term.rstrip(), qualifier
     else:
-        return True, term.rstrip()
+        return True, term.rstrip(), qualifier
 
 
 def save_mapping_rules(mapping_rules: MappingRuleCollection, path: str):
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -110,6 +110,7 @@ from oaklib.utilities.associations.association_differ import AssociationDiffer
 from oaklib.utilities.iterator_utils import chunk
 from oaklib.utilities.kgcl_utilities import generate_change_id
 from oaklib.utilities.lexical.lexical_indexer import (
+    DEFAULT_QUALIFIER,
     add_labels_from_uris,
     apply_synonymizer,
     create_lexical_index,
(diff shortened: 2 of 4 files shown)


src/oaklib/utilities/lexical/lexical_indexer.py
```diff
@@ -124,7 +125,7 @@ def create_lexical_index(
                     term2 = term
                     for tr in pipeline.transformations:
                         if tr.type.code == TransformationType.Synonymization:
-                            synonymized, term2 = apply_transformation(term2, tr)
+                            synonymized, term2, _ = apply_transformation(term2, tr)
```
hrshdhgd, line 128: @cmungall  Question: At this step the returned  `qualifier` value is ignored. If not, would it influence `rel`
```
rel = RelationshipToTerm(
                        predicate=pred,
                        element=curie,
                        element_term=term,
                        pipeline=pipeline.name,
                        synonymized=synonymized,
                    )
```

src/oaklib/utilities/lexical/lexical_indexer.py
```diff
@@ -406,14 +407,18 @@ def apply_transformation(term: str, transformation: LexicalTransformation) -> st
         )
 
 
-def apply_synonymizer(term: str, rules: List[Synonymizer]) -> str:
+def apply_synonymizer(term: str, rules: List[Synonymizer]) -> Tuple:
```
cmungall, line 410: always include as precise types as possible

better:

`Tuple[bool, str, str]`

even better, something like:

`Tuple[bool, str, SYNONYM_QUALIFIER_TYPE]`

have a look at existing tuple declarations in OAK

Also add a docstring here. In general docstrings should go on public methods, and *especially* central ones
like this that are really necessary for understanding synonymizer

src/oaklib/utilities/lexical/lexical_indexer.py
cmungall: This is why I didn't want a special purpose `synonymized`, we end up with a lot of ad-hoc additional attributes. Pipelines and transformations were always adequate for this task.

cmungall (approved):

src/oaklib/utilities/lexical/lexical_indexer.py
hrshdhgd: Done

hrshdhgd:
Please hold on merging this. I still need to test this code.

hrshdhgd:
subject_id | subject_label | predicate_id | object_id | object_label | mapping_justification | mapping_tool | confidence | subject_match_field | object_match_field | match_string | object_preprocessing
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
ABCD:2 | foo bar | skos:narrowMatch | ABCD:1 | other foo bar | semapv:LexicalMatching | oaklib | 0.8 | rdfs:label | oio:hasBroadSynonym | foo bar | semapv:RegularExpressionReplacement

@cmungall @matentzn  Does this make sense? or do I have the `predicate_id` flipped?

matentzn (commented):
The table looks exactly right to me!

matentzn:
@cmungall We just tried for two hours to figure out how the lexmatch "inverted" logic works and I cant wrap my head around it.

```
- preconditions:
      subject_match_field_one_of:
        - oio:hasExactSynonym
        - rdfs:label
      object_match_field_one_of:
        - oio:hasBroadSynonym
    postconditions:
      predicate_id: skos:broadMatch
      weight: 2.0
```

This is the rule.

This is the result:

subject_id | subject_label | predicate_id | object_id | object_label | mapping_justification | mapping_tool | confidence | subject_match_field | object_match_field | match_string | object_preprocessing
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
ABCD:2 | foo bar | skos:narrowMatch | ABCD:1 | other foo bar | semapv:LexicalMatching | oaklib | 0.8 | rdfs:label | oio:hasBroadSynonym | foo bar | semapv:RegularExpressionReplacement


The `subject_match_field` and `object_match_field` exactly match the preconditions of the rule. Yet the predicate is narrowMatch (narrowMatch is "correct" in this case but this is not the point, the point is that it should correspond to the postcondition).

src/oaklib/utilities/lexical/lexical_indexer.py
```diff
@@ -399,21 +411,44 @@ def apply_transformation(term: str, transformation: LexicalTransformation) -> st
...
+            return (False, term, DEFAULT_QUALIFIER)
     else:
         raise NotImplementedError(
             f"Transformation Type {typ} {type(typ)} not implemented {TransformationType.CaseNormalization.text}"
         )
 
 
-def apply_synonymizer(term: str, rules: List[Synonymizer]) -> str:
-    tmp_term = term
+def apply_synonymizer(term: str, rules: List[Synonymizer]) -> Tuple[bool, str, str]:
```
cmungall, line 426: i have a mild preference to name tuples, see for example RELATIONSHIP

src/oaklib/utilities/lexical/lexical_indexer.py
```diff
@@ -399,21 +411,44 @@ def apply_transformation(term: str, transformation: LexicalTransformation) -> st
...
+    """Apply synonymizer rules declared in the given match-rules.yaml file.
+
+    The basic concept is looking for regex in labels and replacing the ones that match
+    with the string passed in 'match.replacement'. Also set qualifier ('match.qualifier')
+    as to whether the replacement is an 'exact', 'broad', 'narrow', or 'related' synonym.
+
+    Note: This function "yields" all intermediate results (for each rule applied)
```
cmungall, line 433: no need for scare quotes!

cmungall (approved):

Issue #329, opened 2022-10-21
`oaklib:cli.annotate()` may need some redesign.


hrshdhgd:
Currently the `annotate` command in the presence of a `--text-file`, iterates through each line and then passes the line to the implementation layer to annotate which in turn `return`s or `yield`s a result.

In your use case for `oakx-oger` , OGER takes input files (various formats) as a whole and spits out output(s) accordingly. OGER does not seem to have a functionality where you pass a word and it gives back its annotation.

Options to address this:

- There could be a flag that indicates if the `--text-file` is passed on to the implementation and let it handle the reading and processing? [default = False]
- (I don't like this) There be an `if` check for `OGERImplementation` which decides to pass the `--text-file` as a param to the `annotate_text()` in `oger_implementation.py`.

any other ideas @cmungall ?

cmungall:
- Add a new method `annotated_text()` with same arguments (other than first) to `annotate` in the interface
 - default implementation should be to read file line by line and pass inputs to `annotate`
 - oger_implementation will override this; this will be the main method
 - oger_implementation will also provide an implementation of `annotate()` that creates a temp file and passes that to oger

PR #333, opened 2022-10-26
Annotate by passing `text-file` to the impl layer

hrshdhgd:
- [x] Added a new param `forward_text_file` to `TextAnnotationCongfiguration`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "oaklib"
-version = "0.1.49"
+version = "0.0.0"
 description = "Ontology Access Kit: Python library for common ontology operations over a variety of backends"
 authors = ["cmungall <cjm@berkeleybop.org>"]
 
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "oaklib"
-version = "0.0.0"
+version = "0.1.49"
```
cmungall, line 3: was this intentional?

src/oaklib/cli.py
```diff
@@ -989,13 +992,17 @@ def term_metadata(terms, reification: bool, output_type: str, output: str):
     "-L",
     help="path to lexical index. This is recreated each time unless --no-recreate is passed",
 )
+@click.option(
+    "--forward-text-file", "-F", help="Implementation level text analysis.", default=False
```
cmungall, line 996: Just `--text-file` is better

cmungall: actually we have text file, so I don't see the point of this

cmungall:
is this to address #329?

pyproject.toml
hrshdhgd: yes for testing purposes ... I intend to roll it back

PR #334, opened 2022-10-26
Split `annotate_text` and `annotate_file`

### LINKED ISSUE #329: `oaklib:cli.annotate()` may need some redesign.

hrshdhgd:
Fixes #329

### DIFF
--- a/src/oaklib/implementations/gilda.py
+++ b/src/oaklib/implementations/gilda.py
@@ -1,6 +1,7 @@
 """A text annotator based on Gilda."""
 
 from dataclasses import dataclass
+from io import TextIOWrapper
 from typing import Iterator
 
 from oaklib.datamodels.text_annotator import TextAnnotation, TextAnnotationConfiguration
@@ -38,3 +39,10 @@ class GildaImplementation(TextAnnotatorInterface):
                 curie=f"{match.term.db}:{match.term.id}",
                 label=match.term.entry_name,
             )
+
+    def annotate_file(
+        self, text_file: TextIOWrapper, configuration: TextAnnotationConfiguration = None
+    ) -> Iterator[TextAnnotation]:
+        for line in text_file.readlines():
+            line = line.strip()
+            self.annotate_text(line, configuration)
--- a/src/oaklib/implementations/ols/ols_implementation.py
+++ b/src/oaklib/implementations/ols/ols_implementation.py
@@ -73,6 +73,9 @@ class BaseOlsImplementation(TextAnnotatorInterface, SearchInterface, MappingProv
     def annotate_text(self, text: str) -> Iterator[TextAnnotation]:
         raise NotImplementedError
 
+    def annotate_file(self, text: str) -> Iterator[TextAnnotation]:
+        raise NotImplementedError
+
     # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     # Implements: OboGraphInterface
     # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
+++ b/src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
@@ -1,6 +1,7 @@
 import logging
 from abc import ABC
 from dataclasses import dataclass, field
+from io import TextIOWrapper
 from typing import Any, ClassVar, Dict, Iterable, Iterator, List, Tuple, Union
 from urllib.parse import quote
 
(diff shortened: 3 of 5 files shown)


hrshdhgd:
I have tested this with [`oakx-oger`](https://github.com/hrshdhgd/oakx-oger) and things look promising.

src/oaklib/implementations/gilda.py
```diff
@@ -25,6 +26,13 @@ class GildaImplementation(TextAnnotatorInterface):
     def annotate_text(
         self, text: TEXT, configuration: TextAnnotationConfiguration = None
     ) -> Iterator[TextAnnotation]:
+        """
+        Annotate a piece of text.
+
+        :param text: Text to be annotated.
+        :param configuration: Text annotation configuration.
+        :return: A generator function that returns annotated results.
```
cmungall, line 34: I often say "yields..." for methods that return iterators

src/oaklib/implementations/gilda.py
```diff
@@ -38,3 +46,16 @@ def annotate_text(
                 curie=f"{match.term.db}:{match.term.id}",
                 label=match.term.entry_name,
             )
+
+    def annotate_file(
+        self, text_file: TextIOWrapper, configuration: TextAnnotationConfiguration = None
+    ) -> Iterator[TextAnnotation]:
+        """Annotate text in a file.
+
+        :param text_file: Text file that is iterated line-by-line.
```
cmungall, line 55: say: processed line by line

src/oaklib/implementations/gilda.py
```diff
@@ -38,3 +46,16 @@ def annotate_text(
...
+        :param configuration: Text annotation configuration.
+        :return: result of `annotate_test()`
```
cmungall, line 57: it's always better to say nothing than to say something with zero information content :-)

and it's better to say nothing than say something wrong :-)

src/oaklib/implementations/ols/ols_implementation.py
```diff
@@ -73,6 +73,9 @@ def labels(self, curies: Iterable[CURIE]) -> Iterable[Tuple[CURIE, str]]:
     def annotate_text(self, text: str) -> Iterator[TextAnnotation]:
         raise NotImplementedError
 
+    def annotate_file(self, text: str) -> Iterator[TextAnnotation]:
```
cmungall, line 76: shouldn't be necessary as calling this will raise NotImplementedError via the default implementation... but not a big deal

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
```diff
@@ -126,6 +127,17 @@ def labels(self, curies: Iterable[CURIE], allow_none=True) -> Iterable[Tuple[CUR
     def annotate_text(
         self, text: str, configuration: TextAnnotationConfiguration = None
     ) -> Iterator[TextAnnotation]:
+        """
```
cmungall, line 130: note this is repetitious with the interface. Big danger of things getting out of sync.

we need to decide on a general pattern for documenting implementations. Thus far I have omitted docstrings. But if we do include them they should add information and not repeat.

We could say something like "implements annotate_text from text_annotator by calling the /annotate/ endpoint using ontoportal client"

src/oaklib/implementations/ontoportal/ontoportal_implementation_base.py
```diff
@@ -139,6 +151,19 @@ def annotate_text(
         results = self._get_json("/annotator", params=params)
         return self._annotator_json_to_results(results, text, configuration)
 
+    def annotate_file(
+        self, text_file: TextIOWrapper, configuration: TextAnnotationConfiguration = None
+    ) -> Iterator[TextAnnotation]:
+        """Annotate text in a file.
```
cmungall, line 157: see comments above

src/oaklib/implementations/gilda.py
```diff
@@ -38,3 +46,16 @@ def annotate_text(
                 curie=f"{match.term.db}:{match.term.id}",
                 label=match.term.entry_name,
             )
+
+    def annotate_file(
```
cmungall, line 50: you don't need to repeat this code. it's inherited from the interface

src/oaklib/interfaces/text_annotator_interface.py
```diff
@@ -92,3 +93,16 @@ def annotate_text(
                             match_string=text[ix : ix + len(k)],
                         )
                         yield ann
+
+    def annotate_file(
```
cmungall, line 97: this is perfect

but note you don't have to repeat this in all implementations. Only override this if you don't want the default implementation

e.g an oger implementation may require a text file, so this is the right place to intercept

cmungall (changes requested):

cmungall (approved):

PR #335, opened 2022-10-27
`fill_table` now handles comments in tsv file.

### LINKED ISSUE #313: `fill-table` should respect commented rows
I'm passing a SSSOM tsv file as a table to be filled for a resource (owl, obo or db file). It works fine when I remove the `metadata` and `prefix_map` from the file and just pass the `dataframe` part of the tsv file. When I pass the tsv as is i.e. `metadata` + `prefix_map` + `dataframe`, it gets stuck while parsing the column names of the `dataframe` (which is parsed as a list) and says it expects a string.

e.g.
when I run this in MONDO:
```runoak -i tmp/mondo-ingest.db fill-table tmp/mondo.sssom.tsv  ```

it throws, 
```
../oaklib/utilities/table_filler.py", line 126, in parse_table
    m = LIST_PATTERN.match(v)
TypeError: expected string or bytes-like object
```

See https://github.com/monarch-initiative/mondo/pull/5491 and also https://github.com/monarch-initiative/mondo/issues/5329#issuecomment-1275384941

cc: @matentzn

### PR DESCRIPTION
**Pull request by hrshdhgd**
Fixes #313

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ authors = ["cmungall <cjm@berkeleybop.org>"]
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.12"
+python = ">=3.9,<4.0.0"
 curies = "^0.4.0"
 nxontology = "^0.4.0"
 pronto = "^2.5.0"
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -3714,6 +3714,12 @@ def diff_via_mappings(
     help="Delimiter between columns in input and output",
 )
 @click.option(
+    "--comment",
+    default="#",
+    show_default=True,
+    help="Comment indicator at the beginning of a row.",
+)
+@click.option(
     "--relation",
     multiple=True,
     help="Serialized YAML string corresponding to a normalized relation between two columns",
@@ -3729,6 +3735,7 @@ def fill_table(
     table_file,
     output,
     delimiter,
+    comment,
     missing_value_token,
     allow_missing: bool,
     relation: tuple,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: src/oaklib/cli.py [hunk 1]
```diff
@@ -3713,6 +3713,12 @@ def diff_via_mappings(
     show_default=True,
     help="Delimiter between columns in input and output",
 )
+@click.option(
+    "--comment",
```
cmungall, line 3717: hmm, I prefer something more descriptive but I guess this is consistent with pandas

src/oaklib/cli.py
```diff
@@ -3820,6 +3827,9 @@ def fill_table(
     - TODO: use most optimized method for whichever backend
     """
     tf = TableFiller(settings.impl)
+    with open(table_file) as input_file:
+        comment_lines = [x for x in input_file.readlines() if x.startswith(comment)]
```
cmungall, line 3831: this is a bit inefficient if it's a large table and we know comments are only at the top

src/oaklib/cli.py
hrshdhgd: Initially i thought of `--comment-indicator` but as you mentioned, went with the pandas param

pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = ["cmungall <cjm@berkeleybop.org>"]
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.12"
+python = ">=3.9,<4.0.0"
```
cmungall, line 10: was this intentional?

cmungall (approved):

PR #345, opened 2022-11-06
Windows fixes (make test suite pass on windows)

dalito:
This PR includes several fixes to make the test suite pass on windows. It also adds Windows to the CI tests.

csv-files were loaded to sqlite by shell commands which do not work in the typical Windows environment. I replaced this by using pandas for reading csv and writing to sqlite. pandas was already an indirect dependency (via sssom) so I thought it is OK to use. Because the old code had links to stackoverflow that indicated the importance of bulk-uploading of large files, I added the option to load csv in chunks. 

So far I have kept the old function `sqlite_bulk_load` in `src/oaklib/implementations/sqldb/sqlite_utils.py` but its code is gone and it calls the new `sqlite_bulk_load2`. However, if the function is not used elsewhere we could replace the old one entirely by the new function `sqlite_bulk_load2`. What do you suggest, should we keep backwards compatibility or only use the new function?

### DIFF
--- a/tests/test_utilities/test_sqlite_utils.py
+++ b/tests/test_utilities/test_sqlite_utils.py
@@ -1,7 +1,10 @@
 import sqlite3
 import unittest
 
-from oaklib.implementations.sqldb.sqlite_utils import sqlite_bulk_load
+from oaklib.implementations.sqldb.sqlite_utils import (
+    sqlite_bulk_load,
+    sqlite_bulk_load2,
+)
 from tests import INPUT_DIR, OUTPUT_DIR
 
 TSV = INPUT_DIR / "foo.tsv"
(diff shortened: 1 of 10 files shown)


tests/test_utilities/test_sqlite_utils.py
```diff
@@ -13,7 +16,17 @@ class TestSqliteUtils(unittest.TestCase):
...
+        sqlite_bulk_load(DB, TSV, TBL_NAME, cols=["a", "b", "c"])
+        con = sqlite3.connect(str(DB))
+        rows = list(con.execute(f"SELECT * FROM {TBL_NAME}"))
+        self.assertGreater(len(rows), 5)
+        self.assertIn(("MGI", "MGI:1918911", "0610005C13Rik"), rows)
+
+    def test_chunked_bulkload(self):
+        if DB.exists():
+            DB.unlink()
+        args = dict(chunksize=1000, sep="\t", comment="!", names=list("abc"))
```
cmungall, line 28: I believe for pandas this will remove *any* text that comes after a `!`, not just a line that starts with this.

cmungall (approved):
Thanks so much @dalito!

I would like to have the ability to bulkload in the most performant way possible, and I am not sure if pandas is the way to do that. However, I think it's best to start with a cross-platform implementation and then add optimizations from there.

I would like to explore removing pandas dependencies from sssom as well, but as you say in the interim we are not adding any more dependencies for users here. We can explore refactoring strategies later.

tests/test_utilities/test_sqlite_utils.py
dalito: "!" is only ignoring the rest of the line but we should make sure that the first data line is not used as column names. I improved the test. https://github.com/INCATools/ontology-access-kit/pull/347

PR #353, opened 2022-11-15
Passing a exclusion list to `annotate` CLI

### LINKED ISSUE #349: Pass exclusion list to lexmatch and annotate
E.g. brand names for pesticides in chebi often result in false positive matches

hrshdhgd:
- [x] `token_exclusion_list` added to `TextAnnotationConfiguration` for `annotate()` CLI
Addresses #349

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "oaklib"
-version = "0.1.0"
+version = "0.0.0"
 description = "Ontology Access Kit: Python library for common ontology operations over a variety of backends"
 authors = ["cmungall <cjm@berkeleybop.org>"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "oaklib"
-version = "0.0.0"
+version = "0.1.0"
 description = "Ontology Access Kit: Python library for common ontology operations over a variety of backends"
 authors = ["cmungall <cjm@berkeleybop.org>"]
 
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -1107,8 +1107,7 @@ def annotate(
             #         ann.subject_source = line
             #         writer.emit(ann)
         else:
-            words = tuple(x for x in words if x not in terms_to_remove)
-            text = " ".join(words)
+            text = " ".join(tuple(x for x in words if x not in terms_to_remove))
             for ann in impl.annotate_text(text, configuration):
                 writer.emit(ann)
     else:
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -1045,7 +1045,7 @@ def annotate(
     output: str,
     lexical_index_file: str,
     matches_whole_text: bool,
-    exclude_terms:str,
+    exclude_terms: str,
     text_file: TextIO,
     output_type: str,
 ):
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -1032,6 +1032,12 @@ def term_metadata(terms, reification: bool, output_type: str, output: str):
     "-L",
     help="path to lexical index. This is recreated each time unless --no-recreate is passed",
 )
+@click.option(
+    "--exclude-terms",
+    "-x",
+    multiple=True,
+    help="Text file or list of terms to exclude from annotation. Each newline separated entry is a distinct text.",
+)
 @output_option
 @output_type_option
 def annotate(
@@ -1039,6 +1045,7 @@ def annotate(
     output: str,
     lexical_index_file: str,
     matches_whole_text: bool,
+    exclude_terms:str,
     text_file: TextIO,
     output_type: str,
 ):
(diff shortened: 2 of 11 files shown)


src/oaklib/cli.py
```diff
@@ -1072,6 +1079,13 @@ def annotate(
     impl = settings.impl
     writer = _get_writer(output_type, impl, StreamingYamlWriter, datamodels.text_annotator)
     writer.output = output
+    terms_to_remove = []
```
cmungall, line 1082: consider putting this into a utility function, to avoid repeating the same chunk of code

src/oaklib/cli.py
```diff
@@ -1072,6 +1079,13 @@ def annotate(
...
+    if len(exclude_terms) == 1 and Path(exclude_terms[0]).exists():
+        with open(exclude_terms[0]) as f:
+            terms_to_remove = f.read().splitlines()
+    else:
+        terms_to_remove = list(exclude_terms)
```
cmungall, line 1087: use consistent terminology. I think `token_exclusion_list` communicates what you are doing here. In ontologies people use term to mean the whole concept, not words in the label

src/oaklib/cli.py
```diff
@@ -1093,7 +1107,7 @@ def annotate(
             #         ann.subject_source = line
             #         writer.emit(ann)
         else:
-            text = " ".join(words)
+            text = " ".join(tuple(x for x in words if x not in terms_to_remove))
```
cmungall, line 1110: there is a strange asymmetry here. if a file is passed, it's left to the implementation to apply filter logic. If it's passed as strings then the logic is in the CLI

src/oaklib/interfaces/text_annotator_interface.py
```diff
@@ -95,14 +95,19 @@ def annotate_text(
                         yield ann
 
     def annotate_file(
-        self, text_file: TextIOWrapper, configuration: TextAnnotationConfiguration = None
+        self,
+        text_file: TextIOWrapper,
+        terms_to_remove: List[str],
```
cmungall, line 100: let's not extend the signature of the method, this is not sustainable, if we keep doing this it will be a big ad-hoc set of key-values. Instead, extend the configuration data model

src/oaklib/cli.py
```diff
@@ -3489,6 +3508,14 @@ def lexmatch(output, recreate, rules_file, lexical_index_file, add_labels, terms
...
     else:
         ruleset = None
+
+    if exclude_terms:
+        terms_to_remove = []
+        if len(exclude_terms) == 1 and Path(exclude_terms[0]).exists():
+            with open(exclude_terms[0]) as f:
+                terms_to_remove = f.read().splitlines()
+        else:
+            terms_to_remove = list(exclude_terms)
```
cmungall, line 3518: it looks like nothing is done with this

src/oaklib/cli.py
```diff
@@ -1032,13 +1032,20 @@ def term_metadata(terms, reification: bool, output_type: str, output: str):
     "-L",
     help="path to lexical index. This is recreated each time unless --no-recreate is passed",
 )
+@click.option(
+    "--exclude-terms",
+    "-x",
+    multiple=True,
+    help="Text file or list of terms to exclude from annotation. Each newline separated entry is a distinct text.",
```
cmungall, line 1039: I suggest `--exclude-tokens` and for docs "file or list of tokens to filter from input prior to annotation. If passed as text file, each newline separated entry is a distinct token"

hrshdhgd: Done

src/oaklib/cli.py
hrshdhgd: Yeah I was going to start with `lexmatch` but realized `annotate` has an immediate need for implementation. I've commented the code out for now.

src/oaklib/cli.py
hrshdhgd: This was just for prototyping and testing. Fixed it. Everything is consolidated to the implementation (`annotate_text`).

src/oaklib/cli.py
hrshdhgd: Done!

src/oaklib/cli.py
hrshdhgd: Done.

src/oaklib/interfaces/text_annotator_interface.py
hrshdhgd: When you say configuration data model, what does that mean?

hrshdhgd: Makes sense, sorry for the oversight.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "oaklib"
-version = "0.0.0"
+version = "0.1.0"
```
cmungall, line 3: error?

src/oaklib/utilities/ner_utilities.py
```diff
@@ -0,0 +1,22 @@
+"""
+Utilites used for annotation and lexmatch processes.
+"""
+
+from pathlib import Path
+from typing import List
+
+
+def get_exclusion_token_list(exclude_tokens: str) -> List[str]:
```
cmungall, line 9: may be good to also have it accept Path

cmungall: `Union[str, Path]`

cmungall (approved):

pyproject.toml
hrshdhgd: Fixed

PR #358, opened 2022-11-22
issue 357

### LINKED ISSUE #357: Enforce range of dcterms:creator to be a human with an orcid (using https)
Currently many ontologies have different practice for the range of dcterms:creator and contributor

Even ontologies that attempt to follow good practice frequently use the "wrong" URI, using the http not the https URI for an orcid. While this is fine for web browsing, for linked data we need to use the same URIs, and by convention the https orcid is the semantic one to use.

This will also facilitate joins with orcid.owl https://raw.githubusercontent.com/cthoyt/wikidata-orcid-ontology/main/orcid.owl

Action: add constraints to the omo schema to restrict to the correct URI

cmungall:
- Adding orcid constraint to creator and contributor, fixes #357
- Extending tests to cover contributor/orcid case.
- using structured pattern
- Adding pattern matching to validation.


src/oaklib/datamodels/ontology_metadata.yaml
```diff
@@ -720,15 +735,24 @@ slots:
     is_a: provenance_property
     slot_uri: dcterms:contributor
     multivalued: true
-    range: Thing
+    range: HomoSapiens
```
matentzn, line 738: I don't think this is the best way to provide a range for the slot. The contributor does not have to be human - it should be possible to express arbitrary agents here, in particular RORs, or wikidata identifiers for people that don't want an orcid..

PR #375, opened 2022-12-02
Adding additional documentation for conversion to FHIR.

cmungall:
- Adding additional documentation for conversion to FHIR
- Adding ability to search by mapping ID (e.g xref)

### DIFF
--- a/docs/converters/index.rst
+++ b/docs/converters/index.rst
@@ -1,0 +1,36 @@
+.. _converters:
+
+Converters
+===============
+
+OAK uses a number of different Data Models (See  :ref:`datamodels`) for representing ontologies and
+ontology-related information.
+
+Command Line Usage
+-------------------
+
+The general pattern is:
+
+.. code-block:: bash
+
+    $ runoak -i INPUT-SPEC dump -O FORMAT -o OUTPUT-FILE
+
+Converting from Obo Graphs
+---------------------------
+
+Currently the only converters implemented are for converting between
+the OboGraphs data model and other models.
+
+.. code-block:: python
+
+        >>> from oaklib.utilities.obograph_utils import load_obograph
+        >>> graph = load_obograph("hp.obo.json")
+
+.. toctree::
+   :maxdepth: 3
+   :caption: Contents:
+
+   data-model-converter
+   obo-graph-to-fhir
+   obo-graph-to-obo-format
+   obo-graph-to-owl
--- a/src/oaklib/converters/data_model_converter.py
+++ b/src/oaklib/converters/data_model_converter.py
@@ -10,10 +10,16 @@ from oaklib.types import CURIE
 
 @dataclass(eq=False)
 class DataModelConverter(ABC):
-    """Base class for all inter-data model converters."""
+    """Base class for all inter-data model converters.
+
+    Do not use this directly: use one of the subclasses.
+    """
 
     curie_converter: curies.Converter = None
+    """Converts between CURIEs and URIs"""
+
     labeler: Callable[[CURIE], Optional[str]] = None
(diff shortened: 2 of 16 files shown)


docs/converters/index.rst
```diff
@@ -0,0 +1,36 @@
...
+ontology-related information.
+
+Command Line Usage
+-------------------
+
+The general pattern is:
+
+.. code-block:: bash
+
+    $ runoak -i INPUT-SPEC dump -O FORMAT -o OUTPUT-FILE
```
joeflack4, line 16: Was this  not in the docs before? Glad it's there now~

docs/converters/index.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Converting from Obo Graphs
+---------------------------
+
+Currently the only converters implemented are for converting between
```
joeflack4, line 21: Converter to obograph might be nice at some point, but not sure if there is a need.

joeflack4: And certainly we'll want direct converters without needing the obograph intermediary.

src/oaklib/converters/data_model_converter.py
```diff
@@ -10,10 +10,16 @@
...
+    """Base class for all inter-data model converters.
+
+    Do not use this directly: use one of the subclasses.
+    """
 
     curie_converter: curies.Converter = None
+    """Converts between CURIEs and URIs"""
+
     labeler: Callable[[CURIE], Optional[str]] = None
+    """A function that returns a label for a given CURIE"""
```
joeflack4, line 22: Codestyle: I guess this is a docstring and not a normal comment `#` because it is a `Callable`

src/oaklib/converters/obo_graph_to_fhir_converter.py
```diff
@@ -46,11 +46,45 @@
 
 @dataclass
 class OboGraphToFHIRConverter(DataModelConverter):
-    """Converts from OboGraph to FHIR."""
+    """Converts from OboGraph to FHIR.
+
+    - An ontology is mapped to a FHIR `CodeSystem <https://build.fhir.org/codesystem.html>`_.
+    - Each node in the OboGraph is converted to a FHIR Concept.
+    - Each CURIE/URI in the OboGraph is treated as a CURIE when it becomes a code (e.g. "HP:0000001")
```
joeflack4, line 53: @cmungall I imagine we will be in discussion on this for a little bit. In most situations, the consensus that I got from the FHIR docs, FHIR RDF group, and others, is that we should not use CURIEs and use something like [FHIR Coding](https://build.fhir.org/datatypes.html#Coding)s.

src/oaklib/converters/obo_graph_to_fhir_converter.py
```diff
@@ -46,11 +46,45 @@
 
 @dataclass
 class OboGraphToFHIRConverter(DataModelConverter):
-    """Converts from OboGraph to FHIR."""
+    """Converts from OboGraph to FHIR.
```
joeflack4, line 49: Nice docs here. Thanks!

src/oaklib/datamodels/fhir.yaml
```diff
@@ -137,4 +141,68 @@ classes:
       version:
       code:
       display:
-      userSelected:
\ No newline at end of file
+      userSelected:
+
+  ConceptMap:
```
joeflack4, line 146: Got more models here I see, nice

joeflack4 (commented):
@cmungall I know you merged already but here are my comments. Only some minor changes I suggested. Figuring out when/where to use CURIEs by the Jan Connectathon would be good, but we'll _hopefully_ get some feedback from that event either way.

PR #379, opened 2022-12-05
yield results from a generator fixes  #376

### LINKED ISSUE #376: --annotate-text does not work with bioportal for all but first line
```python
def annotate_file(
        self,
        text_file: TextIOWrapper,
        configuration: TextAnnotationConfiguration = None,
    ) -> Iterator[TextAnnotation]:
...
...
      return self.annotate_text(line, configuration)
```

there is no way for this to run over subsequent lines

See: https://realpython.com/lessons/looping-over-iterables/

Also:

 - annotate text throws error if `--annotate-text` is used and file has an empty line
 - there are various other typing issues

hrshdhgd:
When a function yields results and this function is called another function that is also intended to yield results of an iterable, the calling function uses `yields from` This is offered by python versions > 3.3. This is called [generator delegation](https://docs.python.org/3/whatsnew/3.3.html).

Fixes #376

### DIFF
--- a/src/oaklib/datamodels/search_datamodel.py
+++ b/src/oaklib/datamodels/search_datamodel.py
@@ -11,23 +11,39 @@
 # license: https://creativecommons.org/publicdomain/zero/1.0/
 
 import dataclasses
-import sys
 import re
-from jsonasobj2 import JsonObj, as_dict
-from typing import Optional, List, Union, Dict, ClassVar, Any
+import sys
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


src/oaklib/datamodels/search_datamodel.py
```diff
@@ -11,23 +11,39 @@
 # license: https://creativecommons.org/publicdomain/zero/1.0/
 
 import dataclasses
-import sys
```
cmungall, line 14: did you mean to include commits on this file? it's not related to this PR?

PR #395, opened 2022-12-08
Pass `prefix_map` to `lexmatch` output and clean the same before generating SSSOM output

### LINKED ISSUE #350: `meta` as a param for `lexmatch`
When we `lexmatch` through the [CLI](https://github.com/INCATools/ontology-access-kit/blob/18e5ddd6f591d0d01f0cccd2ba6c2173b86dd183/src/oaklib/cli.py#L3433), it calls the [`lexical_index_to_sssom`](https://github.com/INCATools/ontology-access-kit/blob/18e5ddd6f591d0d01f0cccd2ba6c2173b86dd183/src/oaklib/cli.py#L3492) which in turn uses [`meta`](https://github.com/INCATools/ontology-access-kit/blob/18e5ddd6f591d0d01f0cccd2ba6c2173b86dd183/src/oaklib/utilities/lexical/lexical_indexer.py#L205) as a parameter. In the absence of the `meta` param, it uses a [default one from sssom](https://github.com/INCATools/ontology-access-kit/blob/18e5ddd6f591d0d01f0cccd2ba6c2173b86dd183/src/oaklib/utilities/lexical/lexical_indexer.py#L238).

Would it make sense for lexmatch to have the metadata parameter as well?

hrshdhgd:
- [x] Passes prefix_map from oak to`lexmatch` output.
- [x] This avoid the mapping file from having a huge noisy `prefix_map` as a header of the SSSOM tsv file.

Fixes #350

### DIFF
--- a/src/oaklib/utilities/lexical/lexical_indexer.py
+++ b/src/oaklib/utilities/lexical/lexical_indexer.py
@@ -203,6 +203,7 @@ def lexical_index_to_sssom(
     lexical_index: LexicalIndex,
     ruleset: MappingRuleCollection = None,
     meta: Metadata = None,
+    prefix_map: dict = None,
     subjects: Collection[CURIE] = None,
     objects: Collection[CURIE] = None,
     symmetric: bool = False,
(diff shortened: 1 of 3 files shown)


hrshdhgd:
@cmungall: Should we take out `meta` as a parameter for `lexical_index_to_sssom()` now that ew pass a `prefix_map` through?

src/oaklib/utilities/lexical/lexical_indexer.py
```diff
@@ -203,6 +203,7 @@ def lexical_index_to_sssom(
     lexical_index: LexicalIndex,
     ruleset: MappingRuleCollection = None,
     meta: Metadata = None,
```
hrshdhgd, line 205: Do we need this?

cmungall (approved):

PR #419, opened 2022-12-20
FHIR: CodeSystem.property

joeflack4:
Updates
```
FHIR
- Add: unit tests for specific ontologies
- Add: unit test helpers to download and use cache
- Update: LinkML: CodeSystemProperty
- Update: OboGraph to FHIR Converter: Now populates CodeSystem.property
```

### DIFF
--- a/src/oaklib/converters/obo_graph_to_fhir_converter.py
+++ b/src/oaklib/converters/obo_graph_to_fhir_converter.py
@@ -14,6 +14,7 @@ from linkml_runtime.dumpers import json_dumper
 from oaklib.converters.data_model_converter import DataModelConverter
 from oaklib.datamodels.fhir import (
     CodeSystem,
+    CodeSystemProperty,
     Coding,
     Concept,
     ConceptDesignation,
@@ -155,6 +156,8 @@ class OboGraphToFHIRConverter(DataModelConverter):
         target.id = source.id
         edges_by_subject = index_graph_edges_by_subject(source)
         logging.info(f"Converting graph to obo: {source.id}, nodes={len(source.nodes)}")
+        self.predicates_to_export = set()
+        # CodeSystem.concept
         for n in source.nodes:
             logging.debug(f"Converting node {n.id}")
             self._convert_node(
@@ -163,6 +166,16 @@ class OboGraphToFHIRConverter(DataModelConverter):
                 target=target,
                 include_all_predicates=include_all_predicates,
             )
+        # CodeSystem.property
+        # todo's
+        #  i. code: mostly URIs, which don't conform to [^\s]+(\s[^\s]+)* (https://hl7.org/fhir/datatypes.html#code)
+        #  ii. description: can get, but tedious; downloading and caching and looking up in source ontologies
+        #  iii. type: ideally Coding (https://build.fhir.org/datatypes.html#Coding). The property value is a code
+        #  defined in an external code system. This may be used for translations, but is not the intent.
+        #  https://hl7.org/fhir/codesystem-concept-property-type.htm
+        target.property = [
+            CodeSystemProperty(code=x, uri=x, type="code") for x in self.predicates_to_export
+        ]
         return target
 
     def _convert_node(
@@ -172,6 +185,7 @@ class OboGraphToFHIRConverter(DataModelConverter):
         target: CodeSystem,
         include_all_predicates: bool = True,
     ) -> Concept:
+        """Converts a node to a FHIR Concept. Also collects predicates to be included in CodeSystem.property."""
         _id = self.code(source.id)
         logging.debug(f"Converting node {_id} from {source}")
         concept = Concept(code=_id, display=source.lbl)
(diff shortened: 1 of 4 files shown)


tests/test_converters/test_obo_graph_to_fhir.py
```diff
@@ -10,28 +13,74 @@
...
 
-ONT = INPUT_DIR / "go-nucleus.json"
-OUT = OUTPUT_DIR / "go-nucleus.fhir.codesystem.json"
-
 
 class OboGraphToFHIRTest(unittest.TestCase):
-    """Tests OBO JSON -> FHIR."""
+    """Tests OBO JSON -> FHIR. Different ontologies have unique structures, so test some specifics for those."""
+
+    def _load_ontology(self, url: str, download_path: str, use_cache: bool = True) -> GraphDocument:
```
joeflack4, line 20: Added some stable URLs for reproducible tests.

**Downloads of obographs**
I hesitated on this partially because I don't know if you will like the way this is right now. If we can change OAK a little though so that it is fast to download the source OWL, convert it to Obographs or sqlite, and then run the FHIR converter, that would make this simpler.

**Caching downloads**
Also, I highly prefer the caching as it speeds up development and running of the tests. If there's a different way we want to do this (e.g. pystow), or if you don't prefer caching, we can change. Or if you prefer, I can write these tests within our [TIMS server automation repo](https://github.com/HOT-Ecosystem/tims-ts/pull/1).

tests/test_converters/test_obo_graph_to_fhir.py
```diff
@@ -10,28 +13,74 @@
...
+        cs: CodeSystem = json_loader.load(str(out), target_class=CodeSystem)
         self.assertEqual("CodeSystem", cs.resourceType)
         [nucleus_concept] = [c for c in cs.concept if c.code == NUCLEUS]
         self.assertEqual("nucleus", nucleus_concept.display)
         self.assertTrue(nucleus_concept.definition.startswith("A membrane-bounded organelle"))
         parents = [x for x in nucleus_concept.property if x.code == "parent"]
         self.assertEqual(len(parents), 1)
         self.assertTrue(parents[0].valueCode == IMBO)
+
+    def test_convert_mondo(self):
```
joeflack4, line 52: I think I'll need to test cases specific to each ontology, so felt like a method for each one was best.

tests/test_converters/test_obo_graph_to_fhir.py
```diff
@@ -10,28 +13,74 @@
...
-    def test_convert(self):
-        """Tests parsing then converting to fhir documents."""
-        gd: GraphDocument = json_loader.load(str(ONT), target_class=GraphDocument)
-        self.converter.dump(gd, OUT, include_all_predicates=True)
-        cs: CodeSystem
-        cs = json_loader.load(str(OUT), target_class=CodeSystem)
+    def test_convert_go_nucleus(self):
+        """General test & specifis for go nucleus."""
+        ont = INPUT_DIR / "go-nucleus.json"
+        out = OUTPUT_DIR / "CodeSystem-go-nucleus.json"
```
joeflack4, line 40: FYI: `CodeSystem-<SYSTEM_NAME>.json` is supposedly the FHIR standard for the filename.

src/oaklib/datamodels/fhir.yaml
```diff
@@ -128,10 +128,18 @@ classes:
   CodeSystemProperty:
```
joeflack4, line 128: @cmungall In the future if I need to edit the LinkML yaml, do I need to manually re convert the dataclass? I think I can find how to do that in the LinkML docs, but just want to check and also see if there's any special considerations  / steps I should take when doing this in OAK.

src/oaklib/converters/obo_graph_to_fhir_converter.py
```diff
@@ -163,6 +165,11 @@ def _convert_graph(
                 target=target,
                 include_all_predicates=include_all_predicates,
             )
+        # CodeSystem.property
```
joeflack4, line 168: Added `CodeSytem.property` conversion

tests/test_converters/test_obo_graph_to_fhir.py
```diff
@@ -10,28 +14,85 @@
 from tests import IMBO, INPUT_DIR, NUCLEUS, OUTPUT_DIR
 from tests.test_implementations import ComplianceTester
 
-ONT = INPUT_DIR / "go-nucleus.json"
-OUT = OUTPUT_DIR / "go-nucleus.fhir.codesystem.json"
+
+DOWNLOAD_TESTS_ON = False
```
joeflack4, line 18: @cmungall I think these ontologies I'm downloading average about 200mb. I figured it would not be desirable to slow down the test suite in general by doing these downloads all the time, so I added `DOWNLOAD_TESTS_ON = False`. So, the purpose of these tests is to run on an ad hoc basis.

I could commit a subset of these ontologies and that way they can always be run, but I want to focus first on getting some more content before the Connectathon first if that's alright.

joeflack4:
@cmungall I'm working through `flake8` / `black` checks right now, but otherwise this is ready for you to review.

I basically updated `CodeSystemProperty` in the LinkML data model and updated the Obographs to FHIR converter.

I'm going to continue working on FHIR updates a lot the next couple weeks. But I'll leave this PR as it is now (smaller), and use it as a base branch for my next FHIR PR coming soon.

src/oaklib/datamodels/fhir.yaml
cmungall: Yes. This is easy. Like most of our repos, everything is coordinated by the Makefile. If you edit the yaml, just make the python target

tests/test_converters/test_obo_graph_to_fhir.py
cmungall: I think this is fine for now

In general it's better to have the control be via environment variable

We need better solutions in general in OAK for integration tests

cmungall (approved):

joeflack4:
I saw your responses, thanks! I see this is approved now, so I am merging.

PR #424, opened 2022-12-23
Maintains no-banana zone.

### LINKED ISSUE #423: gilda makes bananas
```
✗ runoak -i gilda: annotate -W "Atypical Antipsychotic Agents"
object_id: CHEBI:CHEBI:65191
object_label: second generation antipsychotic
subject_end: 29
subject_start: 1
```

not sure if this is gilda itself or the way we wrap it - either way, OAK is a no-banana zone!

hrshdhgd:
Fixes #423

### DIFF
--- a/src/oaklib/implementations/gilda.py
+++ b/src/oaklib/implementations/gilda.py
@@ -42,8 +42,8 @@ class GildaImplementation(TextAnnotatorInterface):
 
         for match in gilda.ground(text):
             term_id_split = match.term.id.split(":")
-            if len(term_id_split) == 1: 
-                curie=f"{match.term.db}:{match.term.id}"
+            if len(term_id_split) == 1:
+                curie = f"{match.term.db}:{match.term.id}"
             elif len(term_id_split) == 2:
                 curie = match.term.id
             else:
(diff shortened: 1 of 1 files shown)


hrshdhgd:
> See latest comments in the issue - the PR needs to be more conservative/defensive
> 
> IF there are 2 `:`s in the final CURIE
> 
> * if it's a banana, remove the first
> * if it's not (e.g. `ABC:DEF:123`) then we should at least emit a warning, possibly even raise an error

Have a fix in place

cmungall (approved):

src/oaklib/implementations/gilda.py
```diff
@@ -41,8 +42,20 @@ def annotate_text(
         import gilda
 
         for match in gilda.ground(text):
+            term_id_split = match.term.id.split(":")
```
cthoyt, line 45: It's highly advisable to use the Bioregistry for doing these kinds of normalizations rather than re-implementing it again and again which could potentially introduce bugs.

cmungall: I think the fix is for gilda to always return CURIEs?

cmungall: @cthoyt - did you see the parent issue? #423

PR #425, opened 2022-12-23
SciSpacy plugin compatibility

hrshdhgd:
- [x] Added `model` as `TextAnnotationConfiguration` attribute.
 - [x] Update `annotate()` to receive `scispacy` model names for NLP. Here's the list:

Available SciSpacy models
 1. `en_ner_craft_md`: A spaCy NER model trained on the CRAFT corpus.
 2. `en_ner_jnlpba_md`: A spaCy NER model trained on the JNLPBA corpus.
 3. `en_ner_bc5cdr_md`: A spaCy NER model trained on the BC5CDR corpus.
 4. `en_ner_bionlp13cg_md`: A spaCy NER model trained on the BIONLP13CG corpus.
 5. `en_core_sci_scibert`: A full spaCy pipeline for biomedical data with  a ~785k vocabulary and allenai/scibert-base as the transformer model.
 6. `en_core_sci_sm`: A full spaCy pipeline for biomedical data.
 7. `en_core_sci_md`: A full spaCy pipeline for biomedical data with a larger vocabulary and 50k word vectors.
 8. `en_core_sci_lg`: A full spaCy pipeline for biomedical data with a larger vocabulary and 600k word vectors.

Avaliable Spacy Models: English pipelines optimized for CPU.
1. `en_core_web_sm`: Components: tok2vec, tagger, parser, senter, ner, attribute_ruler, lemmatizer.
2. `en_core_web_md`: Components: tok2vec, tagger, parser, senter, ner, attribute_ruler, lemmatizer.
3. `en_core_web_lg`: Components: tok2vec, tagger, parser, senter, ner, attribute_ruler, lemmatizer.
4. `en_core_web_trf`: Components: transformer, tagger, parser, ner, attribute_ruler, lemmatizer.

EntityLinkers:
1. `umls`: Links to the Unified Medical Language System, levels 0,1,2 and 9. This has ~3M concepts.
2. `mesh`: Links to the Medical Subject Headings. 
3. `rxnorm`: Links to the RxNorm ontology. RxNorm contains ~100k concepts focused on normalized names for clinical drugs.
4. `go`: Links to the Gene Ontology. The Gene Ontology contains ~67k concepts focused on the functions of genes.
5. `hpo`: Links to the Human Phenotype Ontology.


cmungall (changes requested):
I think you need to update the PR description?

In this particular case, I prefer extending TextAnnotatorConfig to include a `model` parameter  with a range of string. Having a named pre-trained model is not specific to SciSpacy, any annotator that uses ML rather than string matching is using some pre-trained model.

In future we may still want a way of passing in genuinely plugin-specific parameters, but this can be done with a generic key-value list, rather than passing in a filename string that the plugin is expected to parse

cmungall:
I'm also reading https://github.com/allenai/scispacy/issues/463

From the perspective of OAK, if I request annotation using HP (e.g. `-i scispacy:hp`) then I want HP IDs back.

The simplest thing to do in the short term is to simply ignore the linker and treat scispacy purely as a UMLS annotator. The user can then do their own mapping and filtering (e.g. using the MappingProviderInterface and existing implementations like the SRI NodeNormalizer, see https://github.com/INCATools/ontology-access-kit/releases/tag/v0.1.61)

We *could* have the scispacy plugin also do the mapping, such that if I said `scispacy:hp` it would use the hpo linker, and then behind the scenes map the UMLS IDs back to HPO IDs. But I think this makes the plugin more complicated as it would have to rely on some additional service, and doesn't ultimately buy us much.

Note later on we will extend the annotator framework to take in arbitrary value sets (see https://github.com/INCATools/ontology-access-kit/releases/tag/v0.1.58), and we can revisit using the linker here.

src/oaklib/cli.py
```diff
@@ -1330,6 +1337,11 @@ def annotate(
         if exclude_tokens:
             token_exclusion_list = get_exclusion_token_list(exclude_tokens)
             configuration.token_exclusion_list = token_exclusion_list
+        if model:
+            configuration.model = model
+        # if plugin_config:
```
cmungall, line 1342: no point keeping around

cmungall (approved):

PR #426, opened 2022-12-30
More FHIR updates

joeflack4:
Updates
```
commit 0fe7a343c52975259d8e944e291a1c9ef4c89614 (HEAD -> fhir-updates-2)
    Obograph to FHIR updates
    - use_curies_foreign_concepts: default changed to True
    - test cases: updated release URIs
    - bugfix: IDs needed to be formatted in a different way

commit accb465957e2cd92cc3da5ac0fdd118508a2cb4c
    FHIR
    - Update: Changed concept references from CURIEs to codes (for native concepts) or URIs (for foreign concepts) by default.
    - Update: Removed @type appearing in CodeSystem output
    - Add: param: code_system_url
    - Add: param: code_system_id

    FHIR work-in-progress:
    - Add: param use_curies_native_concepts to allow for CURIEs for native concept references (WIP)
    - Add: param use_curies_foreign_concepts to allow for CURIEs for foreign concept references (WIP)
    - Add: param native_uri_stems to allow user to list URI stems. That is, for URIs of native concepts, they can be known to be native if their URI starts with one of these URI strings. (WIP)
    - Add: Unit tests: for RxNorm, SO (WIP)
```

### DIFF
--- a/docs/howtos/fhir-conversions.rst
+++ b/docs/howtos/fhir-conversions.rst
@@ -1,0 +1,64 @@
+FHIR Conversions
+========================
+
+CLI
+===============
+The general pattern is from the CLI is:
+
+.. code-block:: bash
+
+    $ runoak -i INPUT-SPEC dump -O FORMAT -o OUTPUT-FILE
+
+The `fhirjson` format includes a --config-file / -c parameter for its additional parameters.
+
+Example:
+
+.. code-block:: bash
+
+    $ runoak --i go-nucleus.obo dump -o CodeSystem-go-nucleus.json -O fhirjson -c tests/input/fhir_config_example.json
+
+Config file
+===============
+
+Options:  
+
+code_system_id: _The code system ID to use for identification on the server uploaded to.
+See: https://hl7.org/fhir/resource-definitions.html#Resource.id
+
+code_system_url: _Canonical URL for the code system.
+See: https://hl7.org/fhir/codesystem-definitions.html#CodeSystem.url
+
+native_uri_stems: _A comma-separated list of URI stems that will be used to determine whether a
+concept is native to the CodeSystem. For example, for OMIM, the following URI stems are native:
+https://omim.org/entry/, https://omim.org/phenotypicSeries/PS
+
+include_all_predicates (default=True): Include the maximal amount of predicates. Changes the default behavior from
+only exporting: IS_A
+
+use_curies_native_concepts (default=False): _FHIR conventionally uses codes for references to
+concepts that are native to a given CodeSystem. With this option, references will be CURIEs instead.
+
+use_curies_foreign_concepts (default=False): _Typical FHIR CodeSystems do not contain any
+concepts that are not native to that CodeSystem. In cases where they do appear, this converter defaults to URIs for
+references, unless this flag is present, in which case the converter will attempt to construct CURIEs.
+
+predicate_period_replacement (default=False): _Predicates URIs populated into `CodeSystem.
+concept.property.code` and `CodeSystem.concept.property.code`, but the popular FHIR server "HAPI" has a bug in whih
+periods '.' cause errors. If this flag is present, periods will be replaced with underscores '_'.
+
+Example:
+
+
+.. code-block:: json
+
+    {
+        "code_system_id": "HPO",
+        "code_system_url": "http://purl.obolibrary.org/obo/hp.owl",
+            "native_uri_stems": [
+                "http://purl.obolibrary.org/obo/HP_"
+            ],
+        "include_all_predicates": true,
+        "use_curies_native_concepts": false,
+        "use_curies_foreign_concepts": false,
+        "predicate_period_replacement": true
+    }
--- a/docs/howtos/fhir-conversions.rst
+++ b/docs/howtos/fhir-conversions.rst
@@ -1,15 +1,37 @@
 FHIR Conversions
-========================
+================
+
+:term:`FHIR` (Fast Healthcare Interoperability Resources) is a standard for exchanging healthcare data.
+
+OAK includes methods for converting ontologies or ontology fragments to FHIR CodeSystems.
+However, there is no agreed-upon standard way of converting ontologies to FHIR, because
+each ontology has its own conventions and metadata vocabularies, and FHIR makes some
+assumptions that don't hold for ontologies in OBO such as all terms belonging to
+the same code system.
+
+For this reason we make FHIR conversion highly configurable to suit the needs
+of a particular project or FHIR server.
 
 CLI
-===============
-The general pattern is from the CLI is:
+---
+
+Here we will use the OAK ``dump`` command to export an entire ontology to FHIR.
+
+The general pattern for the ``dump`` command is from the CLI is:
 
 .. code-block:: bash
 
-    $ runoak -i INPUT-SPEC dump -O FORMAT -o OUTPUT-FILE
+    $ runoak -i INPUT dump -O FORMAT -o OUTPUT-FILE
+
+For dumping to the FHIR json serialization, use ``fhirjson`` as the format:
+
+.. code-block:: bash
+
+    $ runoak -i INPUT dump -O fhirjson -o OUTPUT-FILE
+
 
 The `fhirjson` format includes a --config-file / -c parameter for its additional parameters.
+
 
 Example:
 
@@ -18,7 +40,7 @@ Example:
     $ runoak --i go-nucleus.obo dump -o CodeSystem-go-nucleus.json -O fhirjson -c tests/input/fhir_config_example.json
 
 Config file
-===============
+-----------
 
 Options:  
 
--- a/docs/howtos/fhir-conversions.rst
+++ b/docs/howtos/fhir-conversions.rst
@@ -42,6 +42,8 @@ Example:
 Config file
 -----------
 
+See :ref:`obo_graph_to_fhir_converter` for the full list of options.
+
 Options:  
 
 code_system_id: _The code system ID to use for identification on the server uploaded to.
--- a/src/oaklib/cli.py
+++ b/src/oaklib/cli.py
@@ -2300,7 +2300,7 @@ def descendants(
     code_system_url: For `fhirjson` only. Canonical URL for the code system.
     See: https://hl7.org/fhir/codesystem-definitions.html#CodeSystem.url
 
-    native_uri_stems: For `fhirjson` only. A comma-separated list of URI stems that will be used to determine whether a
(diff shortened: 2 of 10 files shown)


src/oaklib/cli.py
```diff
@@ -1922,6 +1922,27 @@ def descendants(terms, predicates, display: str, output_type: str, output: TextI
     show_default=True,
     help="For formats that export only IS_A by default, this will include all possible predicates",
 )
+@click.option(
+    "--use-curies-native-concepts/--no-use-curies-native-concepts",
```
joeflack4, line 1926: I could probably use some improvement on the descriptions for these CLI options, but for now, here's what I've got:

`--use-curies-native-concepts`
> For FHIR. FHIR conventionally uses codes for references to concepts that are native to a given CodeSystem. With this option, references will be CURIEs instead.

`--use-curies-foreign-concepts`
> For FHIR. Typical FHIR CodeSystems do not contain any concepts that are not native to that CodeSystem. In cases where they do appear, this converter defaults to URIs for references, unless this flag is present, in which case the converter will attempt to construct CURIEs.

`--native-uri-stems`
> For FHIR. A comma-separated list of URI stems that will be used to determine whether a concept is native to the CodeSystem. For example, for OMIM, the following URI stems are native: https://omim.org/entry/,https://omim.org/phenotypicSeries/PS

joeflack4:
@cmungall Re-running test/lint and will fix more if needed.

Any chance this could be merged soon? I released https://github.com/hot-Ecosystem/owl-on-fhir which TIMS depends on, but it won't work until this PR is merged.

FYI that package is partly TIMS devops, but moreso right now just makes it easier to convert OWL to FHIR. Eventually OAK will be a smoother experience for FHIR and this package may become deprecated. I think right now it's missing the ability to do OWL (or Obograph) to FHIR fully from the CLI rather than Python (memory is a little fuzzy as it's been a few months).

src/oaklib/cli.py
cmungall: unfortunately the CLI will blow up if we expose all possible configurations of all possible systems

What I would recommend here is that the CLI accepts as argument a generic configuration file or object, and specific converters can then define their own fields here

Note the DumperInterface already looks like this:

```
    def dump(
        self, path: str = None, syntax: str = None, enforce_canonical_ordering=False, **kwargs
    ):
        """
        Exports current contents.

        :param path: Path to file to write to. If None, then write to stdout.
        :param syntax: Syntax to use. If None, then use the default syntax.
        :param enforce_canonical_ordering:
        :param kwargs: Additional arguments to pass to the dumper
        :return:

...
...
        converter.dump(ogdoc, target=path, **kwargs)
```

so the CLI could be made to accept a yaml file that is just flattened into `**kwargs`

joeflack4: Thanks for the feedback, Chris. I am just unsure of how you want me to implement, I can think of several options. See [Slack](https://obo-communitygroup.slack.com/archives/D046Y7BPEJH/p1680199597627229).

docs/howtos/fhir-conversions.rst
```diff
@@ -0,0 +1,64 @@
+FHIR Conversions
```
joeflack4, line 1: **Docs**
@cmungall I wanted some docs that were easily accessible / in the main docs rather than the code, but I couldn't find a good spot, so I chose here. The main thing I wanted to document was the config file. I was originally going to do `docs/converters/obo-graph-to-fhir.rst`, but the config file isn't actually used by that converter; it's used by `dump()` in CLI.py.

src/oaklib/cli.py
```diff
@@ -2286,23 +2287,65 @@ def descendants(
...
 @click.argument("terms", nargs=-1)
 @click.option("-o", "--output", help="Path to output file")
+@output_type_option
 @click.option(
-    "--include-all-predicates/--no-include-all-predicates",
-    default=False,
-    show_default=True,
-    help="For formats that export only IS_A by default, this will include all possible predicates",
+    "-c",
+    "--config-file",
```
joeflack4, line 2293: **Docs**
I also documented this in `dump()` within `cli.py` So that there's an easy way to know what to do if you're interrogating the `--help` in the command line. But if this is too much, feel free to delete this text, and we can simply direct / link to `docs/howtos/fhir-conversions.rst` or refer them to the docstring for `OboGraphToFHIRConverter` (which doesn't say anything about a config file, but does show params).

src/oaklib/cli.py
```diff
@@ -2330,6 +2368,8 @@ def dump(terms, output, output_type: str, **kwargs):
     impl = settings.impl
     if isinstance(impl, BasicOntologyInterface):
         logging.info(f"Out={output} syntax={output_type}")
+        with open(config_file) as file:
+            kwargs |= json.load(file)
```
joeflack4, line 2372: FYI: Config file implementation; combines any params that appear with `kwargs`.

src/oaklib/converters/obo_graph_to_fhir_converter.py
```diff
@@ -97,47 +99,93 @@ def dump(
...
 
-        By default, only IS_A predicates are converted to ConceptProperties. To override this,
-        specify ``include_all_predicates=True``.
-
-        :param source:
-        :param target:
-        :param include_all_predicates: include the maximal amount of predicates
+        :param source: Source serialization.
+        :param target: Target serialization.
+        :param code_system_id: The code system ID to use for identification on the server uploaded to.
```
joeflack4, line 116: **Docs**
Added documentation for params to `OboGraphToFHIRConverter`.

@cmungall You mentioned
> But for now the SOT docs can be the docstring in the fhir converter class

I'm not sure what SOT means; I tried googling it but couldn't find.

joeflack4 (commented):
@cmungall Assuming all checks pass, this is ready for review again.

src/oaklib/converters/obo_graph_to_fhir_converter.py
cmungall: https://en.wikipedia.org/wiki/Single_source_of_truth

tests/test_converters/test_obo_graph_to_fhir.py
```diff
@@ -63,41 +85,114 @@ def test_convert_go_nucleus(self):
...
+                code_system_id=filename.replace("CodeSystem-", ""),
+                code_system_url="https://github.com/"
+                "loinc/comp-loinc/releases/latest/download/merged_reasoned_loinc.owl",
+                native_uri_stems=["https://loinc.org/"],
+            )
+            self.assertGreater(len(cs.concept), 5000)
             prop_uris: List[str] = [p.uri for p in cs.property]
             self.assertIn("https://loinc.org/hasComponent", prop_uris)
+
+    def test_convert_rxnorm(self):
```
cmungall, line 152: this test downloads a 0.25G file - I don't think this should be a unit test

src/oaklib/cli.py
```diff
@@ -2286,23 +2287,65 @@ def descendants(
...
+    help="""Config file for additional params. Presently used by  `fhirjson` only.
+
+    Options:
+    code_system_id: For `fhirjson` only. The code system ID to use for identification on the server uploaded to.
+    See: https://hl7.org/fhir/resource-definitions.html#Resource.id
+
+    code_system_url: For `fhirjson` only. Canonical URL for the code system.
+    See: https://hl7.org/fhir/codesystem-definitions.html#CodeSystem.url
+
+    native_uri_stems: For `fhirjson` only. A comma-separated list of URI stems that will be used to determine whether a
```
cmungall, line 2303: As you are documenting either a python structure or json structure, where the expected value is 

```json
"native_uri_stems": ["stem1", ... , "stemN"]
```

I don't think this should say comma-separated, unless you intend

```json
"native_uri_stems": "stem1, ..., stemN"
```

src/oaklib/converters/obo_graph_to_fhir_converter.py
```diff
@@ -61,15 +66,12 @@ class OboGraphToFHIRConverter(DataModelConverter):
 
          - TODO: make this configurable
 
-    - Each edge in the OboGraph is converted to a _FHIR ConceptProperty_ if the `include_all_predicates` param is True.
-      Otherwise, will only convert edges if the predicate is in the `DIRECT_PREDICATE_MAP`.
+    - Each edge in the OboGraph is converted to a _FHIR ConceptProperty_ if the `include_all_predicates` param is
+      True. Otherwise, will only convert edges if the predicate is in the `DIRECT_PREDICATE_MAP`.
     - Each synonym in the OboGraph is converted to a _FHIR ConceptDesignation_.
 
         - The synonym predicate is mapped to a _FHIR Coding_, using the `SCOPE_MAP`.
```
cmungall, line 71: I don't think this is correct

docs/howtos/fhir-conversions.rst
```diff
@@ -0,0 +1,64 @@
...
+========================
+
+CLI
+===============
```
cmungall, line 5: you are already using `===` above, should switch to `---`

docs/howtos/fhir-conversions.rst
```diff
@@ -0,0 +1,64 @@
...
+.. code-block:: bash
+
+    $ runoak --i go-nucleus.obo dump -o CodeSystem-go-nucleus.json -O fhirjson -c tests/input/fhir_config_example.json
+
+Config file
+===============
+
+Options:  
+
+code_system_id: _The code system ID to use for identification on the server uploaded to.
```
cmungall, line 25: I like how you added a how-to but I don't think you need to restate everything here. The role of the how-to is to provide additional explanation about why something is done.

See https://diataxis.fr/

src/oaklib/converters/obo_graph_to_fhir_converter.py
```diff
@@ -97,47 +99,93 @@ def dump(
...
-        specify ``include_all_predicates=True``.
-
-        :param source:
-        :param target:
-        :param include_all_predicates: include the maximal amount of predicates
+        :param source: Source serialization.
+        :param target: Target serialization.
+        :param code_system_id: The code system ID to use for identification on the server uploaded to.
+        See: https://hl7.org/fhir/resource-definitions.html#Resource.id
+        :param code_system_url: Canonical URL for the code system.
```
cmungall: These need to be properly formatted, see https://stackoverflow.com/questions/32566466/multi-line-description-of-a-parameter-description-in-python-docstring

currently results in:

image

cmungall:
Thanks @joeflack4, this should be very valuable for this broader clinical community!

PR #431, opened 2023-01-13
`alias_map` & `synonym_map` flags to `text_annotator`

hrshdhgd:
The idea is having 2 flags (default: False):
 - `--alias-map/--no-alias-map`
 - `--synonym-map/--no-synonym-map`

When either flag is `True`, the corresponding information will show in the output.


src/oaklib/cli.py
```diff
@@ -1320,6 +1327,26 @@ def annotate(
...
         https://incatools.github.io/ontology-access-kit/interfaces/text-annotator
+
+    Aliases can be listed in the output by setting the flag
+    --include-aliases to `true` (default: false).
+
+    Example (using the plugin oakx-spacy):
+
+        runoak -i spacy:sqlite:obo:bero annotate Myeloid derived suppressor cells. --include-aliases
+
+    will yield:
```
cmungall, line 1338: remember https://github.com/INCATools/ontology-access-kit/issues/310

cmungall (approved):

PR #445, opened 2023-02-02
Prohibit sqla 1.4.46 and greenlet 2.0.2 as this was causing problems for some installs

cmungall:
See

- #446
- https://github.com/mapping-commons/sssom-py/issues/343


cmungall:
seems no longer necessary...

sierra-moxon:
@cmungall - do you know what the fix was here?  I just had to apply the same change in order to have both linkml-runtime and oaklib in the same poetry environment. 
this was the error I got:
```
  Command '['/Users/Documents/src/go-fastapi/.venv/bin/python', '-m', 'pip', 'install', '--disable-pip-version-check', '--prefix', '/Users/Documents/src/go-fastapi/.venv', '--no-deps', '/Users/Library/Caches/pypoetry/artifacts/1d/31/1a/756ea27ca1be1c4724695fa553d879c193c73f93f02530e1daa7030055/greenlet-2.0.2-cp39-cp39-macosx_11_0_x86_64.whl']' returned non-zero exit status 1.
```

sierra-moxon (approved):

cmungall:
I think if greenlet releases a whl this should be fixed:

- https://github.com/python-greenlet/greenlet/issues/350

cmungall:
I am told this magic incantation works:

`poetry run pip install greenlet`

seems weird to do a pip install via poetry but if it works...

PR #451, opened 2023-02-14
Add `rustsim` to dependencies; add implementation stub

caufieldjh:
Use rustsim (https://github.com/Knowledge-Graph-Hub/rustsim) for semantic similarity comparisons.
Results should be approximately identical to stock OAK semsim but faster.

This PR is just for including the rustsim dependency and adding the basic functionality.


caufieldjh:
Test fails for 3.10 (on [5051f19](https://github.com/INCATools/ontology-access-kit/pull/451/commits/5051f19442da7e077d4f5cbf7972839ab154d5d2)) because we don't (currently) publish a 3.10 wheel.

pyproject.toml
```diff
@@ -25,15 +25,15 @@ kgcl-schema = "^0.3.1"
...
-pystow = "^0.5.0"
+pystow = "^0.4.4"
 class-resolver = "^0.3.10"
 ontoportal-client = "0.0.3"
 bioregistry = "^0.5.64"
 prefixmaps = "^0.1.2"
 ols-client = "^0.1.1"
 pandas = "^1.5.1"
 linkml-renderer = "^0.1.2"
-airium = "^0.2.5"
```
cmungall, line 36: was there a reason for this?

pyproject.toml
```diff
@@ -25,15 +25,15 @@ kgcl-schema = "^0.3.1"
 funowl = "^0.1.12"
 gilda = {version = "^0.10.1", optional = true}
 kgcl-rdflib = "^0.3.0"
-pystow = "^0.5.0"
+pystow = "^0.4.4"
```
cmungall, line 28: ideally we would not do this

caufieldjh: Reverted in 9e38b39c9106440256fe1ad5407bd3c4187d0338

pyproject.toml
caufieldjh: Reverted in 9e38b39c9106440256fe1ad5407bd3c4187d0338

caufieldjh: I don't know why that happened.

caufieldjh:
Windows tests will fail because we don't build a rustsim windows wheel - see https://github.com/Knowledge-Graph-Hub/rustsim/issues/16

caufieldjh:
Move rustsim to Implementations from Interfaces, following design patterns in this project

cmungall:
I assume this is subsumed into #488 now?

caufieldjh:
Yes, let's consider this a redundant PR

PR #453, opened 2023-02-18
Fix non-stopping CI on windows

dalito:
The bug can be seen here: https://github.com/INCATools/ontology-access-kit/actions/runs/4209345211/jobs/7306256452

Same fix as in https://github.com/linkml/linkml/pull/1202


dalito:
This is ready now!

cmungall (approved):

PR #455, opened 2023-02-23
Add subcommand to produce association rollup report

pkalita-lbl:
This introduces a `rollup` subcommand for producing an association rollup report. It uses the existing `AssociationProviderInterface.associations()` method for parsing and querying associations. It also adds some new elements to `datamodels/associations.yaml` to represent the data structure that is formatted into the report (you can also get this data structured dumped as JSON or YAML).

A few things to consider:

- I'm not really sure about the "rollup" nomenclature. I don't know if there is a more widely recognized term than that.
- The specification of `--object-group` option is a little ad-hoc. The way it is now specifically limits the user to 2 levels of grouping depth. We talked about another possible approach of accepting one flat list and have the subcommand assemble the terms into a tree. This opens the possibility of arbitrary levels of nested groups. The underlying data structure supports that; the biggest hurdle is probably just thinking about how the report should handle that. Let me know if you think we should explore that direction more.
- The report is kind of intentionally bare-bones at the moment, but I'm open to suggestions.


cmungall:
I like rollup, the older nomenclature is map2slim

I like keeping things open for both simple 1 level as well as more levels, and the possibility of simple TSV or yaml reports, but only if this doesn't add too much complexity

cmungall (approved):

cmungall:
it works!

Just making a few additional commits (probably deserve to be on their own PR...)...

PR #456, opened 2023-02-27
Adding a query() method and a query command to the CLI.

cmungall:
This allows passing through queries to the specific adapter that is being used;
i.e SQL for sqlite, SPARQL for ubergraph, ontobee, rdf files. As such
this is inherently non-portable, but may still be useful for advanced use cases.

An ubergraph notebook is also included.

This PR also normalizes the URIs for the LinkML datamodels.


cmungall:
there will be a notebook later (the ubergraph notebook shows the equivalent for sparql)

PR #468, opened 2023-03-14
update bioregistry dependency to fix downstream dependency errors

sierra-moxon:
see: https://github.com/INCATools/kgcl/pull/24 
see: https://github.com/biolink/biolink-model-toolkit/issues/113
and finally upstream: https://github.com/biopragmatics/bioregistry/issues/705

Needed to prevent downstream failures on fixed bioregistry error.
A release would be appreciated with this fix as well.


pkalita-lbl (commented):
The `poetry.lock` file is being deleted 😬 ???

hrshdhgd:
Oh no!!! Good catch Patrick. @sierra-moxon , reason?

cmungall (approved):

Issue #470, opened 2023-03-14
Add a generator to NDEx CX format


cmungall:
CX format is used by cytoscape and ndexbio (https://www.ndexbio.org/). It would be useful to be able to export to CX from OAK, via the obograph model.

I envision this in two parts:

1. ensuring the core logical content of the ontology (minimally: node.{id,lbl} and edge.{sub,pred,obj}) is exported
2. exporting the kgviz stylesheets to the cx style language

the first should be pretty straightforward, at the core cx is very similar to obographs

```json
    "nodes": [
      {
        "@id": 0,
        "n": "ACTA1",
        "r": "uniprot:P02568"
      },
      {
        "@id": 1,
        "n": "CASP3",
        "r": "uniprot:A8K5M2"
      },
...
  {
    "edges": [
      {
        "@id": 378,
        "s": 8,
        "t": 11,
        "i": "interacts with"
      },
      {
        "@id": 379,
        "s": 11,
        "t": 16,
        "i": "interacts with"
      },
...
```

At first I looked into doing this via a linkml model of CX. However, the structure of CX is a little unusual - the top level structure is a list, and different object types form elements of the list, with some assumptions about order. And CX is also self-describing in that a data dictionary of attributes is passed, so there is no one fixed schema of keys for the json.

I tried just manually constructing dict objects but my initial attempts produced invalid structures (I was using the ndex ui to test, not sure if there is a more efficient way)

On slack @bgyori suggests using [ndex2](https://ndex2.readthedocs.io/en/latest/ndex2.html), I think this makes the most sense, although I am loathe to add new dependencies, especially where it may introduce conflicts over shared libraries such as networkx

cmungall:
A very bad experimental first pass #471

PR #471, opened 2023-03-14
First pass at CX writer, 470

cmungall:
Note: this is almost certainly in the wrong path.

See #470 for discussion


cmungall:
closed in favor of #479

Issue #474, opened 2023-03-16
Add a PHAF parser


cmungall:
Not sure if you want to take this on @manulera 

As a first pass this would only use entity and term fields, but in future the phaf fields could be mapped to a standard data dictionary

This is less necessary now we have #473 and the ability to pass in a simple 2-column format for enrichment

PR #476, opened 2023-03-16
PHAF parser and tests added

manulera:
Addresses #474

Includes the parser and analogous test to that in `tests/test_parsers/test_hpoa_association_parser.py`


cmungall:
one flake8 check to fix!

you can also do a `tox -e flake8` to check locally

cmungall (changes requested):
flake8

manulera:
I think it is fixed now, sorry about that.

cmungall (approved):

PR #478, opened 2023-03-16
add a subset of OAK logos to this repo

nlharris:


cmungall (changes requested):
looks like the logos went in at the top level?

nlharris:
ugh, sorry. i'm not actually sure how to move them from this PR (it won't let me edit them) so maybe we should just merge this and then i'll do another pr to move them?

cmungall:
sounds good!

nlharris:
couldn't find a way to move them via github; had to add them to the correct folder and then, separately, delete from the main folder.

PR #488, opened 2023-03-21
RustSimImplementation

hrshdhgd:
Implements `rustsim` to calculate semantic similarity.


tests/test_implementations/test_rustsim_implementation.py
```diff
@@ -0,0 +1,84 @@
...
+class TestRustSimImplementation(unittest.TestCase):
+    """Implementation tests for Rust-based semantic similarity."""
+
+    def setUp(self) -> None:
+        """Set up"""
+        # Calling get_implementation_from_shorthand() alone fails on Windows
+        try:
+            oi = get_adapter(f"rustsim:sqlite:///{str(DB)}")
+        except FileNotFoundError:
+            oi = SqlImplementation(OntologyResource(slug=f"sqlite:///{str(DB)}"))
```
cmungall, line 24: I don't understand the reasoning here

This should test rustsim. It should fail fast if for some reason it can't connect

tests/test_implementations/test_rustsim_implementation.py
```diff
@@ -0,0 +1,84 @@
...
+        subj_ancs = set(self.oi.ancestors(VACUOLE, predicates=[IS_A, PART_OF]))
+        obj_ancs = set(self.oi.ancestors(ENDOMEMBRANE_SYSTEM, predicates=[IS_A, PART_OF]))
+        jaccard = jaccard_similarity(subj_ancs, obj_ancs)
+        calculated_jaccard = len(subj_ancs.intersection(obj_ancs)) / len(subj_ancs.union(obj_ancs))
+        self.assertAlmostEqual(calculated_jaccard, jaccard)
+
+    def test_rustsim_mrca(self):
+        """Tests Rust implementations of Most Recent Common Ancestor (mrca) with score."""
+        expected_tuple = ("CARO:0000000", 21.05)
+        mrca = mrca_and_score(self.information_content_scores)
```
cmungall, line 48: not testing the interface

caufieldjh:
Discussed with @hrshdhgd and @justaddcoffee:
* `get_resource_from_shorthand` in `selector.py` handles Windows filepaths without issue...as long as we aren't using a descriptor like the one currently in `test_rustsim_implementation`: 
```
"rustsim:sqlite:///{str(DB)}"
```
In that case, a Windows filepath gets parsed incorrectly and we get a FileNotFoundError.
* We can avoid that error by restricting rustsim to *only* using the SqlImplementation, meaning we'll just call it with `rustsim:`
As it is now, the RustsimImplementation already assumes it's going to use the SqlImplementation, as that's how it gets closures and IC scores.
* We can also bypass the issue entirely by calling rustsim with a flag instead of as part of the descriptor.

* For the tests, we need to directly call the RustsimImplementation functions - this means separating out the DB data source. Just provide the test data directly without fully relying on other components.

tests/test_implementations/test_rustsim_implementation.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class TestRustSimImplementation(unittest.TestCase):
+    """Implementation tests for Rust-based semantic similarity."""
+
+    def setUp(self) -> None:
+        """Set up"""
+        # Calling get_implementation_from_shorthand() alone fails on Windows
+        try:
+            oi = get_adapter(f"rustsim:sqlite:///{str(DB)}")
```
dalito, line 22:
```python
>>> from pathlib import Path
>>> f"rustsim:sqlite:///{Path(DB).as_posix()}"
'rustsim:sqlite:///C:/this.db'
```

hrshdhgd: Thanks for the suggestion @dalito ! Unfortunately, that didn't work. The problem lies [here](https://github.com/INCATools/ontology-access-kit/blob/main/src/oaklib/selector.py#L184-L188)

```
        if ":" in descriptor and not os.path.exists(descriptor):
            toks = descriptor.split(":")
            scheme = toks[0]
            resource.scheme = scheme
            rest = ":".join(toks[1:])
```

caufieldjh:
@hrshdhgd I talked with @cmungall about this - will follow up on Slack with some other approaches. 
tl;dr - let's set up the RustsimImplementation so it accepts a table of closures as input whether they're from SqlImplmentation or wherever. Then the closures can be sent to rustsim and it can store them in memory so subsequent calls to similarity functions don't have to retrieve the closures all over again

caufieldjh (approved):
Mergeable! Still needs more specific tests - will open issue.

PR #506, opened 2023-04-01
`simple_obo_parser.py` breaks when a phrase is passed to KGCL for synonym removal.

hrshdhgd:
Command trying to pass: `remove synonym 'whole organism' for FBbt:00000001`

As of now the `_simple_value()` in `simple_obo_parser.py` look like this
```
def _simple_value(self, v) -> str:
        return v.split(" ")[0]
```

The synonym we need to remove is `"whole organism"` and `_simple_value()` return just `"whole` instead of the entire phrase which results in no match.

This PR fixes the issue.


cmungall:
simple_value shouldn't be used with synonym, it should only be used when the expected argument is a node, not a literal

PR #518, opened 2023-04-05
Update SSSOM Minor version in OAK

matentzn:
There was a new minor release, and the ^ symbol prevents upgrading.


matentzn:
Error messages is weird; https://pypi.org/project/sssom-schema/

hrshdhgd (approved):
Ran `poetry update` after you changed the toml file. This should fix the issue.

PR #520, opened 2023-04-06
Definitions should be enclosed in double quotes while updating via KGCL

hrshdhgd:
Earlier iteration had `definitions` in single quotes. This PR adds/updates definitions of classes in double quotes


src/oaklib/implementations/pronto/pronto_implementation.py
```diff
@@ -836,10 +836,16 @@ def apply_patch(
             for syn in t.synonyms:
                 if syn.description == patch.old_value:
                     syn.description = patch.new_value
+        elif isinstance(patch, kgcl.NewTextDefinition):
+            t = self._entity(patch.about_node, strict=True)
+            xrefs = t.definition.xrefs if t.definition else []
+            v = f'"{patch.new_value}"'
```
cmungall, line 842: This isn't correct. For the pronto adapter you are using the pronto data model directly, so don't quote

src/oaklib/implementations/simpleobo/simple_obo_implementation.py
```diff
@@ -795,14 +795,17 @@ def apply_patch(
                 raise ValueError(f"Failed to find synonym {patch.old_value} for {t.id}")
             modified_entities.append(patch.about_node)
         elif isinstance(patch, kgcl.NewTextDefinition):
+            v = f'"{patch.new_value}"'
```
cmungall, line 798: in contrast to pronto this is fine, as with this adapter you are working closer to the level of the ascii...

but where is the trailing `[]`s?

cmungall (changes requested):
see comments re quoting, `[]`s

src/oaklib/implementations/pronto/pronto_implementation.py
hrshdhgd: Isn't this is similar to `simple_obo_implementation`  since this is in `apply_patch()` specific to the `kgcl.NewTextDefinition` object which is what we need?

src/oaklib/implementations/simpleobo/simple_obo_implementation.py
hrshdhgd: It's a `NewTextDefinition`, so if there is a `[]` provided it will be added as long as it is within the double quotes. Or am I misunderstanding the question?

src/oaklib/implementations/simpleobo/simple_obo_parser.py
```diff
@@ -631,6 +631,8 @@ def _dump_stanza(self, stanza: Stanza, file: TextIO):
     def _dump_tag_values(self, tag_values: List[TagValue], file: TextIO):
         for tv in tag_values:
             if tv.tag != "id":
+                if tv.tag == "def" and tv.value.startswith("'"):
```
cmungall, line 634: this part shouldn't be necessary

PR #521, opened 2023-04-07
Stanzas are sorted by `Term` and `Typedef`

hrshdhgd:


src/oaklib/datamodels/vocabulary.py
```diff
@@ -146,6 +146,9 @@
 OWL_VERSION_INFO = "owl:versionInfo"
 OWL_VERSION_IRI = "owl:versionIRI"
 
+OBO_TERM = "Term"
```
cmungall, line 149: these shouldn't go in vocabulary as it's entirely an obo syntax thing - aren't these already in simple obo parser?

src/oaklib/implementations/simpleobo/simple_obo_parser.py
```diff
@@ -596,9 +600,25 @@ def order_stanzas(self) -> None:
 
         Does not change tag-value ordering within stanzas
         """
-        stanzas = self.stanzas.values()
-        stanzas = sorted(stanzas, key=lambda x: x.id)
-        self.stanzas = {s.id: s for s in stanzas}
+        term_stanzas = {
```
cmungall, line 603: there may be more compact ways to write this, but the logic is perfect I think!

cmungall (changes requested):
just the vocabulary change

hrshdhgd:
Done!

cmungall (approved):

PR #537, opened 2023-04-12
Adding more attributes to Synonymizer

hrshdhgd:
New schema for `Synonymizer` looks like this:

```
Synonymizer:
   attributes:
    the_rule:
      description: Description of the rule.
      range: string
    match:
      description: Reg-ex rule to match substrings in labels.
      range: string
    match_scope:
      description: Scope of the reg-ex rule
      range: string
    replacement:
      description: Reg-ex rule to replace substrings in labels
      range: string
    qualifier:
      description: Type of match for the new synonym generated.
      range: string
    prefix:
      description: The rule applies to nodes of a specific prefix.
      range: string
    tests:
      description: Unit tests for each rules.
      range: Test
  
  Test:
   attributes:
    input:
     description: Input string for the rule.
    output:
     description: Output based on the rule.
    prefix:
      description: The prefix that qualifies for the rule.
```


cmungall (commented):
this modifies poetry lock but not pyproject.toml - deliberate?

cmungall (approved):

hrshdhgd:
Yes, `click` needed to be updated which has `pywin32` as a dependency. The change is the addition of the missing `.whl` file information amongst other minors one. That results in the error disappearance.

PR #558, opened 2023-05-01
Slight addition to handle KGCL `RemoveTextDefinition`

hrshdhgd:
Here, I'm replacing the definition entirely to "".

Example command:
```
remove definition for FBbt:00000001
```

Implements https://github.com/INCATools/kgcl/pull/28

### DIFF
--- a/src/oaklib/implementations/simpleobo/simple_obo_implementation.py
+++ b/src/oaklib/implementations/simpleobo/simple_obo_implementation.py
@@ -813,8 +813,12 @@ class SimpleOboImplementation(
             t = self._stanza(patch.about_node, strict=True)
             for tv in t.tag_values:
                 if tv.tag == TAG_DEFINITION:
-                    # The strip() portion is to handle a KGCL bug
-                    tv.replace_quoted_part(patch.new_value.strip("'"))
+                    if patch.new_value:
+                        # The strip() portion is to handle a KGCL bug
+                        tv.replace_quoted_part(patch.new_value.strip("'"))
+                    else:
+                        # This is a remove_definition request
+                        tv.value = ""
         elif isinstance(patch, kgcl.NewSynonym):
             t = self._stanza(patch.about_node, strict=True)
             # Get scope from patch.qualifier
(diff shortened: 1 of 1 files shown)


cmungall (changes requested):
To implement RemoveTextDefinition, add another switch in the code to test if the change instantiations RemoveTextDefinition

hrshdhgd:
> To implement RemoveTextDefinition, add another switch in the code to test if the change instantiations RemoveTextDefinition

Done!

src/oaklib/implementations/simpleobo/simple_obo_implementation.py
```diff
@@ -813,8 +813,12 @@ def apply_patch(
             t = self._stanza(patch.about_node, strict=True)
             for tv in t.tag_values:
                 if tv.tag == TAG_DEFINITION:
-                    # The strip() portion is to handle a KGCL bug
-                    tv.replace_quoted_part(patch.new_value.strip("'"))
+                    if isinstance(patch, kgcl.RemoveTextDefinition):
```
cmungall, line 816: this should be at the same level as the other switches

cmungall (changes requested):

src/oaklib/implementations/simpleobo/simple_obo_implementation.py
hrshdhgd: Done

src/oaklib/implementations/simpleobo/simple_obo_implementation.py
```diff
@@ -809,11 +809,16 @@ def apply_patch(
             t = self._stanza(patch.about_node, strict=True)
             t.add_quoted_tag_value(TAG_DEFINITION, patch.new_value.strip("'"), xrefs=[])
             modified_entities.append(patch.about_node)
+        elif isinstance(patch, kgcl.RemoveTextDefinition):
+            t = self._stanza(patch.about_node, strict=True)
+            for tv in t.tag_values:
+                if tv.tag == TAG_DEFINITION:
+                    # This is a remove_definition request
+                    tv.value = ""
```
cmungall, line 817: the tag-val pair should be properly removed

cmungall (changes requested):
see comments

src/oaklib/implementations/simpleobo/simple_obo_implementation.py
hrshdhgd: Done

cmungall (approved):

PR #559, opened 2023-05-01
Leave keys unsorted when exporting yaml

### LINKED ISSUE #529: KGCL yaml statistics output should be deterministic
Running somthing like

```
runoak --stacktrace -i simpleobo:ontologies/maxo_2022-06-24.obo diff -X simpleobo:ontologies/maxo_2023-03-09.obo --statistics -o stats/maxo_diff.txt.yaml
```

Results in non deterministic serialisation of stats/maxo_diff.txt.yaml. 

image

hrshdhgd:
Fixes #529


cmungall (approved):

cmungall:
I'm OK to merge this anyway as we should always leave the keys unsorted

PR #563, opened 2023-05-09
Replaced `simple_mappings_by_curie` to call `sssom_mappings`

### LINKED ISSUE #564: `lexical_index` did not account for `skos:exactMatch` which needs change in `simple_mappings_by_curie` wrap around `sssom_mappings`
Issue: https://github.com/monarch-initiative/mondo-ingest/issues/272#issuecomment-1534888151

Fix: #563

hrshdhgd:
Fixes #564

### DIFF
--- a/src/oaklib/implementations/sqldb/sql_implementation.py
+++ b/src/oaklib/implementations/sqldb/sql_implementation.py
@@ -35,7 +35,7 @@ from semsql.sqla.semsql import (
     DeprecatedNode,
     Edge,
     EntailedEdge,
-    HasMappingStatement,
+    # HasMappingStatement,
     HasSynonymStatement,
     HasTextDefinitionStatement,
     IriNode,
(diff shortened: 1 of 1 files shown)


src/oaklib/implementations/sqldb/sql_implementation.py
```diff
@@ -1609,30 +1611,31 @@ def sssom_mappings(
...
-                subject_id=row.subject,
-                object_id=row.object,
-                predicate_id=row.predicate,
-                mapping_justification=justification,
-            )
-            inject_mapping_sources(mpg)
-            if source and mpg.subject_source != source and mpg.object_source != source:
-                continue
-            yield mpg
+        if bidirectional:
```
cmungall, line 1614: Sometimes I prefer guard classes here but I think it's good either way

cmungall (approved):
We can add a test later//

PR #577, opened 2023-05-23
Fix slug in benchmark target in Makefile (rustsim: -> semsimian)

justaddcoffee:


cmungall (approved):
I'm not totally sure we should be doing benchmarks in the Makefile in the first place but I assume this is temporary...

justaddcoffee:
> I'm not totally sure we should be doing benchmarks in the Makefile in the first place but I assume this is temporary...

Well, we were not sure where else to put it - it doesn't seem crazy to have a target for benchmarking IMHO. Glad to move it though - where would be a better place?

caufieldjh:
We could also make these optional tests (but definitely optional since they, especially #578, are time-instensive)

cmungall:
I think ultimately these belong in pheval, but for now I would suggest having a separate subfolder with its own Makefile. But OK to merge for now

PR #579, opened 2023-05-24
Implement all_by_all_pairwise_similarity in semsimian implementation

justaddcoffee:
Calls all_by_all_pairwise_similarity in semsimian (which doesn't exist yet, see https://github.com/monarch-initiative/semsimian/pull/51)


src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -113,3 +113,36 @@ def pairwise_similarity(
...
+
+        :param subjects:
+        :param objects:
+        :param predicates:
+        :return:
+        """
+        objects = list(objects)
+
+        for result in self.semsimian.all_by_all_pairwise_similarity(subjects,
+                                                                    objects,
```
cmungall, line 136: add the passthru for min thresholds - otherwise creating N^2 objects on the python side is expensive!

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -113,3 +113,36 @@ def pairwise_similarity(
...
+                                                                    set(predicates)):
+            sim = TermPairwiseSimilarity(
+                subject_id=result.subject_id,
+                object_id=result.object_id,
+                ancestor_id=result.ancestor_id,
+                ancestor_information_content=result.ancestor_information_content,
+            )
+            sim.jaccard_similarity = result.jaccard_similarity
```
cmungall, line 144: any reason not to do in the initialization?

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -113,3 +113,40 @@ def pairwise_similarity(
...
+        """
+        objects = list(objects)
+        all_results = self.semsimian.all_by_all_pairwise_similarity(
+            set(subjects), set(objects), set(predicates)
+        )
+        for term1_key, values in all_results.items():
+            for term2_key, result in values.items():
+                jaccard, resnik, phenodigm_score, ancestor_set = result
+                # TODO: Confirm if this 'if' condition is necessary.
+                if len(ancestor_set) > 0:
```
hrshdhgd, line 141: Is this acceptable?

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -113,3 +113,40 @@ def pairwise_similarity(
...
+                    sim = TermPairwiseSimilarity(
+                        subject_id=term1_key,
+                        object_id=term2_key,
+                        ancestor_id=next(
+                            iter(ancestor_set)
+                        ),  # TODO: Change this: gets first element of the set
```
hrshdhgd, line 147: Need a better way to pass `ancestor_id`

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -116,3 +117,46 @@ def pairwise_similarity(
...
+    def all_by_all_pairwise_similarity(
+        self,
+        subjects: Iterable[CURIE],
+        objects: Iterable[CURIE],
+        predicates: List[PRED_CURIE] = None,
+        min_jaccard_similarity: Optional[float] = None,
+        min_ancestor_information_content: Optional[float] = None,
+    ) -> Iterator[TermPairwiseSimilarity]:
+        """
+        Compute similarity for all combinations of terms in subsets vs all terms in objects
```
cmungall, line 130: no need to duplicate docs from interface in implementation

pyproject.toml
```diff
@@ -37,7 +37,7 @@ pysolr = "^3.9.0"
 eutils = ">=0.6.0"
 requests-cache = "^1.0.1"
 click = "*"
-semsimian = "0.1.14"
+semsimian = ">=0.1.14"
```
cmungall, line 40: definitely correct? so nothing in this PR relies on a later version

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -116,3 +117,46 @@ def pairwise_similarity(
...
+        :param subjects:
+        :param objects:
+        :param predicates:
+        :return:
+        """
+        objects = list(objects)
+        all_results = self.semsimian.all_by_all_pairwise_similarity(
+            set(subjects), set(objects), set(predicates) if predicates else None
+        )
+        for term1_key, values in all_results.items():
```
cmungall, line 141: I don't understand why we're doing python N^2 iteration. Isn't the point of this PR to do the N^2 in Rust?

pyproject.toml
hrshdhgd: This PR relies on the latest version .. 0.1.15

src/oaklib/implementations/semsimian/semsimian_implementation.py
cmungall: Ah I see, the bit above does the all x all, this is just yielding the results. But there is no threshold, so it will do n^2 object creations

pyproject.toml
cmungall: OK I will make a new PR where I fix the pyproject

PR #583, opened 2023-05-30
graph-traversal improvements.

cmungall:
This PR makes graph traversal more efficient, and
unifies how some different adapters make use of
relation graphs and entailed edges.

It adds documentation and examples clarifying the
difference between entailment and traversal methods.

It also adds experimental support for dynamic reasoning
using relation-graph


cmungall:
@matentzn @dosumis @balhoff - could I get your eyes on this piece of documentation in particular

https://github.com/INCATools/ontology-access-kit/blob/graph-efficiency/docs/guide/relationships-and-graphs.rst

PR #597, opened 2023-06-13
Semsimian handles thresholds for resnik and jaccard

### LINKED ISSUE #596: semsim optimization: minimum value thresholds should be passed through to underlying rustsim code
currently we call

```python
all_results = self.semsimian.all_by_all_pairwise_similarity(
            set(subjects), set(objects), set(predicates) if predicates else None
        )
```

then filter

```python
            for term2_key, result in values.items():
                jaccard, resnik, phenodigm_score, ancestor_set = result
                if min_jaccard_similarity is not None and jaccard < min_jaccard_similarity:
                    continue
                if (
                    min_ancestor_information_content is not None
                    and resnik < min_ancestor_information_content
                ):
                    continue
```

but we shouldn't filter in python - the thresholds should be passed through

hrshdhgd:
Fixes #596

This can only be merged once: https://github.com/monarch-initiative/semsimian/pull/66 is merged and a new release of `semsimian` is made.


src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -137,19 +137,16 @@ def all_by_all_pairwise_similarity(
         objects = list(objects)
         logging.info(f"Calculating all-by-all pairwise similarity for {len(objects)} objects")
         all_results = self.semsimian.all_by_all_pairwise_similarity(
-            set(subjects), set(objects), set(predicates) if predicates else None
+            set(subjects),
+            set(objects),
+            min_jaccard_similarity,
```
cmungall, line 142: positional arguments make me very uneasy. What happens when we add more methods, we want to add filters for these. We end up with 10 arguments with high potential for error in aligning arguments...

If rust does not support named arguments I think we need to move towards a config object. This can be done later but is important

cmungall (approved):

src/oaklib/implementations/semsimian/semsimian_implementation.py
hrshdhgd: Rust does support named arguments. Making the change.

PR #600, opened 2023-06-16
More benchmarks for semsimian comparisons

caufieldjh:
Note: will move all benchmarks to Pheval soon.
Part of this PR will be moving the benchmarks out of the Makefile.


cmungall:
is this moved to pheval yet?

caufieldjh:
I don't believe we need this here anymore, yes. Closing.

Issue #602, opened 2023-06-27
Update Gilda implementation to enable custom grounders


caufieldjh:
There's a use case for OntoGPT (https://github.com/monarch-initiative/ontogpt/issues/140) in which having custom grounders for Gilda would be helpful. In brief, it would be useful to import a gene list to use for annotation. In an OAK context, a term list can certainly be used as a search query but I'm not aware of a way to treat it as the input without converting to OWL or the like first.

The current [GildaImplementation ](https://github.com/INCATools/ontology-access-kit/blob/main/src/oaklib/implementations/gilda.py) doesn't support providing config details, and I'm guessing that would be necessary to specify custom annotation sets.

PR #603, opened 2023-06-28
Improve descriptor processing for Gilda implementation

cthoyt:
Closes https://github.com/INCATools/ontology-access-kit/issues/602, related to https://github.com/monarch-initiative/ontogpt/issues/140

This PR does three things:

1. It enables passing custom grounders to the `GildaImplementation`
2. It updates the way that descriptors related to the Gilda grounding/NER system are parsed. Before, all that was possible was to write `gilda:`. Now, this PR enables you to write `gilda:<path to custom index>` where the path is to a gzipped TSV file containing the pre-built Gilda index of text lookups.
3. It enables passing arbitrary keyword arguments to `get_adapter()` that are forwarded to the implementation class on instantiation. This is useful if you want to e.g., pass a custom grounder but you're only able to work with `get_adapter()`

This PR also includes relevant tests to demonstrate this usage with `oaklib.get_adapter()`

## Why

I have colleagues who are using OntoGPT and want to use a custom Gilda index that includes additional ontologies and vocabularies. Because there are many levels of complexities and dependencies, there are no direct ways to pass configuration other than through OAK descriptors. Therefore, we wanted to make the descriptors able to pass this configuration.

## Try Yourself

An example on how the default index is generated can be found here: https://github.com/indralab/gilda/blob/master/gilda/generate_terms.py. There are also several ways to generate custom Gilda indexes, for example you can use PyOBO:

```python
from pyobo.gilda_utils import get_gilda_terms
from gilda.generate_terms import dump_terms

terms = get_gilda_terms("umls")
dump_terms(terms, "my_custom_index.tsv.gz")
```

Then, you might write `gilda:my_custom_index.tsv.gz` as your descriptor for OAK or OntoGPT

## Looking forward

It's not ideal that the only way to pass configuration is through the descriptor string. For example, we might want to specify additional configuration for Gilda, such as a prefix prioritization order. Anything past a single configuration that comes after the colon `:` in the descriptor would then be an issue


cthoyt:
@caufieldjh the tests seem to be failing due to an issue importing part of boto. Could this again be a result of poetry over-specifying the dependencies' versions?

caufieldjh:
Thanks so much @cthoyt !
Checking on those test fails...

caufieldjh:
Just failing Win tests now - maybe related to defining path for termfile

cthoyt:
@caufieldjh thank you so much for looking into this!

src/oaklib/implementations/gilda.py
```diff
@@ -48,8 +48,11 @@ def __post_init__(self):
             # when parsed from a descriptor like ``gilda:<path>` via
             # :func:`get_resource_from_shorthand`. If no <path> was
             # given, then this will default to the default Gilda index
-            self.grounder = Grounder(terms=self.resource.slug)
-
+            try:
+                self.grounder = Grounder(terms=self.resource.slug)
```
cthoyt, line 52: is it possible to have a resource where the slug attribute isn't available?

caufieldjh: Not really an issue with gilda or the grounder, just with how it's being called. It's OK if terms=None, but it doesn't work if self.resource=None since there's no slug. That only breaks the parent classes if you call `annotate_text` on them or something. Shouldn't happen through the usual CLI usage of course.

caufieldjh (approved):

Issue #606, opened 2023-06-29
Similarity with `semsimian` raises an error from `curies`


caufieldjh:
Running the similarity command with semsimian raises `curies.api.DuplicateURIPrefixes` with any parameters:
```
$ runoak -i semsimian:sqlite:obo:bfo similarity -p i,p .all @ .all
curies.api.DuplicateURIPrefixes: Duplicate URI prefixes:

https://icd.who.int/browse10/2019/en#/:
        prefix='ICD10WHO' uri_prefix='https://icd.who.int/browse10/2019/en#/' prefix_synonyms=[] uri_prefix_synonyms=[]
        prefix='icd10who' uri_prefix='https://icd.who.int/browse10/2019/en#/' prefix_synonyms=[] uri_prefix_synonyms=[]
```

Running `runoak -i semsimian:sqlite:obo:bfo similarity --help` does the same.

Running the same command without the semsimian prefix in the input slug runs as expected.

caufieldjh:
With verbose:
```
$ runoak -vvv -i semsimian:sqlite:obo:bfo similarity -p i,p .all @ .all
INFO:root:Setting other_languages=()
INFO:root:Settings = Settings(impl=None, autosave=False, associations_type=None, preferred_language=None, other_languages=())
INFO:root:Wrapping an existing OAK implementation to fetch sqlite:obo:bfo
INFO:root:Locator: obo:bfo
INFO:root:Ensuring gunzipped for https://s3.amazonaws.com/bbop-sqlite/bfo.db.gz
INFO:root:Locator, post-processed: sqlite:////home/harry/.data/oaklib/bfo.db
curies.api.DuplicateURIPrefixes: Duplicate URI prefixes:

https://icd.who.int/browse10/2019/en#/:
        prefix='ICD10WHO' uri_prefix='https://icd.who.int/browse10/2019/en#/' prefix_synonyms=[] uri_prefix_synonyms=[]
        prefix='icd10who' uri_prefix='https://icd.who.int/browse10/2019/en#/' prefix_synonyms=[] uri_prefix_synonyms=[]
```

caufieldjh:
Is it related to this change? https://github.com/INCATools/ontology-access-kit/commit/e8446b58d0ac70a97fb203553a104e07c384cff5

caufieldjh:
Minimal reproduction (again, in python 3.9):
```
>>> from oaklib.implementations.semsimian.semsimian_implementation import SemSimianImplementation
[...]

PR #608, opened 2023-07-05
Fix for #606 - omit lower-case prefixes from SqlImplementation prefixmap if upper-case already present

caufieldjh:
Something strange currently happens with ontologies retrieved as sql db - they contain a single duplicate prefix in their prefixmap, so a `curies` error gets raised. This PR avoids the error by removing duplicate prefixes from that prefixmap, where a duplicate is any lower-case version of an upper-case prefix in the map.

### DIFF
--- a/src/oaklib/implementations/sqldb/sql_implementation.py
+++ b/src/oaklib/implementations/sqldb/sql_implementation.py
@@ -386,11 +386,17 @@ class SqlImplementation(
         return has_view
 
     def prefix_map(self) -> PREFIX_MAP:
+
+        def entity_prefix(curie: CURIE):
+            if ":" in curie:
+                return curie.split(":")[0]
+        
         if self._prefix_map is None:
             self._prefix_map = {row.prefix: row.base for row in self.session.query(Prefix)}
             prefixes = self._prefix_map.keys()
 
             # Check for duplicates, or they may raise an error in curies package
+            # If they are not present in the input then remove them from the prefixmap
             # see https://github.com/INCATools/ontology-access-kit/issues/606
             duplicate_prefixes = {
                 k: v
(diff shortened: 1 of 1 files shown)


cthoyt:
As an alternative, I would suggest trying to deal with the incoming data itself. Why are multiple potentially overlapping prefixes present? `curies` doesn't make any opinions on what the right prefixes are, but it does protect against data that obviously has some issues as we're seeing here.

Dropping part of the prefix map could lead to having prefixes in the data that are no longer defined, so I would like to see a more careful check when dropping the prefix from the prefix map that there aren't any CURIEs using that prefix in the SQL databas.

Another option is to say 'hey, you need to conform to some kind of external prefix standard (e.g., the Bioregsitry) rather than defining your own, which is demonstrably error prone'.

caufieldjh:
This is a workaround because it's a blocker for some semsim work we'd like to get done ASAP. For now, the only offending prefix is `icd10who`, and if this is coming in at some point during the assembly of the db versions of each ontology, then there will be some amount of time before we can fully troubleshoot that. I'd be very happy to know where that prefix is coming from.

> Dropping part of the prefix map could lead to having prefixes in the data that are no longer defined, so I would like to see a more careful check when dropping the prefix from the prefix map that there aren't any CURIEs using that prefix in the SQL database.

That's fair. 

> Another option is to say 'hey, you need to conform to some kind of external prefix standard (e.g., the Bioregsitry) rather than defining your own, which is demonstrably error prone'.

In theory, that's what's already happening, as the default prefixmap here is the `prefixmaps` [OBO context](https://github.com/linkml/prefixmaps/blob/main/src/prefixmaps/data/obo.csv).

cmungall (approved):

cmungall:
I approve within 24hrs the new sqlite databases will be published with the duplicate eliminated

PR #609, opened 2023-07-11
Using semsimian v0.1.18 that writes output directly rather than using `oaklib` to write output

### LINKED ISSUE #619: Add ability to pass a limited closure table to semsimian
phenio.db:

```
sqlite> select count(*) from entailed_edge where predicate ='rdfs:subClassOf';
5675458
sqlite> select count(*) from entailed_edge;
46854391
```

it is useful to have the option to use any closure, but if the application knows in advance the union of all possible query predicates then we reduce the size of the python object that has to be passed to rust by ~10x

hrshdhgd:
- [x] Fixes #619 

### TODO
- [x] Remove rc version anchoring of semsimian (done for testing purposes).


src/oaklib/cli.py
```diff
@@ -2647,13 +2650,24 @@ def similarity_pair(terms, predicates, autolabel: bool, output: TextIO, output_t
     type=float,
     help="Minimum value for information content",
 )
+@click.option(
+    "--embeddings-file",
+    type=click.File(mode="r"),
+    help="file containing embeddings of all necessary nodes.",
```
hrshdhgd, line 2656: Add more documentation.

hrshdhgd: Also maybe rename to `--term-vectors`

hrshdhgd:
Introduce a new CLI param named `--config *.yaml` which would have values for:
 -  `low_memory`
 - `terms_vector` etc.

```yaml
low_memory: true
terms_vector: "abcd_embeddings.tsv"

```

and so on.

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -56,8 +56,24 @@ def __post_init__(self):
...
 
-        spo = [r for r in self.wrapped_adapter.relationships(include_entailed=True)]
-        self.semsimian = Semsimian(spo)
+        self.term_pairwise_similarity_attributes = [
+            attr
+            for attr in vars(TermPairwiseSimilarity)
+            if not any(attr.startswith(s) for s in ["class_", "_"])
+        ]
+
+    def create_semsimian_object(self, predicates: List[PRED_CURIE] = None):
```
hrshdhgd, line 65: Not informative enough. Change name to maybe `create_tps_object()`(?)

hrshdhgd:
Irrelevant now due to #645

PR #631, opened 2023-08-07
Create Semsimian object with concerned predicates only.

hrshdhgd:
- [x] Create semsimian object only with concerned predicates
- [x] Pydantic version < 2
- [x] Semsimian anchored at 0.2.0
- [x] Add in-memory caching for better efficiency
- [x] `termset-similarity` command returns `TermSetPairwiseSimilarity` object


pyproject.toml
```diff
@@ -8,6 +8,7 @@ readme = "README.md"
 
 [tool.poetry.dependencies]
 python = ">=3.9,<4.0.0"
+pydantic = "1.10.12"
```
cmungall, line 11: Let's not do this. OAK is a core library. People are importing OAK in combination with other libraries. There is nothing in OAK that depends on 1.10.12 specifically. However, there is a need to pin < 2 due to transitive dependencies not being pinned correctly

tox.ini
```diff
@@ -25,6 +25,7 @@ deps =
     pytest
     coverage
     semsimian
+    gilda
```
cmungall, line 28: this seems an unrelated change?

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -136,6 +157,8 @@ def all_by_all_pairwise_similarity(
         """
         objects = list(objects)
         logging.info(f"Calculating all-by-all pairwise similarity for {len(objects)} objects")
+        if not hasattr(self, "semsimian") or getattr(self, "semsimian", None) is None:
```
cmungall, line 160: this can be simplified to `if getattr(self, "semsimian", None) is None`

cmungall: or even simpler: `if self.semsimian is None`

cmungall (changes requested):
I don't think the logic is quite right. Consider the case where the client does:

```
sim1 = adapter.pairwise_similarity(x,y, predicates=[IS_A])
sim2 = adapter.pairwise_similarity(x,y, predicates=[IS_A, PART_OF])
<some kind of comparison of the two...>
```

I believe that sem2 will give the same result. It will silently use the same cached filtered closure, and the user will be none the wiser.

If we do thing that the standard use case of semsimian is that you will never want to switch predicate sets in one session (not ideal but acceptable in the short term), then it is vital that the above code will fail fast on the second call and inform the user that the cached object is already frozen on the IS_A-only closure.

However, I think a more standard solution is that you have a cache of semsimian objects keyed by the tuple of the ordered list of predicates (ordered to avoid unnecessary re-computation as [IS_A, PART_OF] is the same as [PART_OF, IS_A]).

e.g. one of the object attributes would be

```
    semsimian_cache: Dict[Tuple, Semsimian] = None
```

Then `create_pairwise_similarity_output_object` would turn `predicates` to a tuple `tuple(sorted(predicates))` and use that as a key.

There is a danger here that the client could exhaust memory by iterating through all combos of 100 predicates in uberon... but I think that is acceptable for now (later on we could have a system that did more active cache management).

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -56,8 +56,27 @@ def __post_init__(self):
...
 
-        spo = [r for r in self.wrapped_adapter.relationships(include_entailed=True)]
-        self.semsimian = Semsimian(spo)
+        self.term_pairwise_similarity_attributes = [
+            attr
+            for attr in vars(TermPairwiseSimilarity)
+            if not any(attr.startswith(s) for s in ["class_", "_"])
+        ]
+
+    def create_pairwise_similarity_output_object(
```
cmungall, line 65: 1. consider making the private
2. consider changing the name to make this reflect what it does. What is a pairwise similarity output object anyway?

I think the correct name for the current functionality is something like `_ensure_semsimian_object_is_set`. "ensure" is a good name for a method with side effects that doesn't return anything about ensures the object is in a particular state.

However, see my other comments. I think a better pattern here is for this method to both cache *and* to return a semsimian object.

E.g. pseudocode:

```
  def get_semsimian_object_for_predicates(predicates: List[PRED_CURIE]) -> Semsimian:
      """Returns a semsimian object for the given predicate set, using cached version if available (if not available, create new cache entry"""
     pred_key = <create key>
     if <key not in cache>:
        <generate object and cache it>
     return <dict>[pred_key]
```

hrshdhgd: Made the changes. Went a step further:
- Used `pystow` to create a dir under the `oaklib` module with the ontology name. So let's say we run the command:
```
runoak -i semsimian:sqlite:obo:bfo similarity -p i BFO:0000040 @ BFO:0000002
```
This creates a file within the `oaklib` pystow cache: `.data/oaklib/bfo/BFO/0000050+rdfs/subClassOf.pickle` which will be read if exists or created if absent. So basically `.data/oaklib/ontology_used_in_slug/concatenated_predicate_key.pickle`. Does this work?

hrshdhgd: Since it is a cache file, I didn't take the extra step of formatting the filename i.e. it has slashes in it.

cmungall: interesting! I'd like @kevinschaper's take on this. I worry a bit about magic persistence of the cache. It will be easy for users to accidentally use a stale cache and not be aware.

Of course this is already a danger with using `sqlite:obo:...` where we use pystow to cache the sqlite db

hrshdhgd: We do have [`cache_clear()`](https://github.com/INCATools/ontology-access-kit/blob/2e4d7e6c07bece59f293d8001e65851c14e8585b/src/oaklib/cli.py#L4927-L4942) in cli.py which should take care of this.

tox.ini
hrshdhgd: tox -e py fails locally if this is absent.

src/oaklib/implementations/semsimian/semsimian_implementation.py
hrshdhgd: `if` condition not needed anymore with caching in place.

cmungall (approved):
I'm nervous about the caching to disk but go ahead and merge when ready we can revisit the strategy later

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -56,8 +59,40 @@ def __post_init__(self):
...
 
-        spo = [r for r in self.wrapped_adapter.relationships(include_entailed=True)]
-        self.semsimian = Semsimian(spo)
+        self.term_pairwise_similarity_attributes = [
+            attr
+            for attr in vars(TermPairwiseSimilarity)
+            if not any(attr.startswith(s) for s in ["class_", "_"])
+        ]
+
+    def _ensure_semsimian_object_is_set(
```
cmungall, line 68: The name doesn't match the logic. What happens if you call this a 2nd time with the same predicates?

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -56,8 +59,40 @@ def __post_init__(self):
...
+        self, predicates: List[PRED_CURIE] = None, attributes: List[str] = None
```
cmungall, line 69: if the output changes based on the attribute list then the attribute list must be in the key - otherwise the cache could become corrupted (potentially silently)

but do you even need to parameterize by attributes? 

better to decide on a fixed set of attributes:

- jaccard
- IC of MRCA
- identifier(s) of MRCA(s)?

kevinschaper:
For my use case, ideally (as in, nice to have, not a requirement) I'd be able to generate the cache ahead of time so that I can mount it in place on a docker container. Maybe a strategy for specifying a cache location or cache file?

src/oaklib/implementations/semsimian/semsimian_implementation.py
sierra-moxon: https://pystow.readthedocs.io/en/latest/api/pystow.ensure.html#pystow.ensure -- not sure if this helps, but I use the force=True flag with pystow to force recreation of any files it stores every time it runs (doesn't solve the problem of when to run it to make the cache).

cmungall (changes requested):

src/oaklib/implementations/semsimian/semsimian_implementation.py
hrshdhgd: This issue still remains as of now.

src/oaklib/implementations/semsimian/semsimian_implementation.py
hrshdhgd: For now we are going with in-memory caching.

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -56,8 +59,40 @@ def __post_init__(self):
...
+    ):
+        """Create a new Semsimian object (in rust) with desired predicates only.
+        This basically creates an object with attributes from TermPairwiseSimilarity class.
+        These are used as columns for the output file generated via `similarity`.
+        :param predicates: List of desired predicates, defaults to None.
+        """
+        predicates.sort() if isinstance(predicates, List) else [predicates].sort()
```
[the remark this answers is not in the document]
hrshdhgd: Solved

pyproject.toml
hrshdhgd: Fixed

src/oaklib/implementations/semsimian/semsimian_implementation.py
hrshdhgd: Resolved

src/oaklib/implementations/semsimian/semsimian_implementation.py
hrshdhgd: If `attributes` is None, on the Semsimian side there already is a default list to fall back on. This list is:
```
            "subject_id",
            "object_id",
            "jaccard_similarity",
            "ancestor_information_content",
            "phenodigm_score",
            "cosine_similarity",
            "ancestor_id"
``` 

This is missing a few and I'm working on adding it.

caufieldjh:
Because changes in this branch are necessary for Monarch use cases and will not be impacted by Windows-specific errors, I'm going to consider this state sufficient for an RC release and will open a fresh PR for the Windows issue.

cmungall (approved):

PR #646, opened 2023-08-28
First pass at interface for association search

cmungall:


hrshdhgd (approved):

tests/test_implementations/test_sqldb.py
```diff
@@ -607,6 +607,41 @@ def test_store_associations(self):
         oi.autosave = True
         self.compliance_tester.test_store_associations(oi)
 
+    def test_associations(self):
+        spec = InputSpecification(
+            ontology_resources={"go": {"selector": str(DB)}},
+            association_resources={"gaf": {"selector": str(INPUT_GAF)}},
```
hrshdhgd, line 613: What are the column names for the `INPUT_GAF` tsv file

cmungall: I think this is a bit orthogonal to this PR, but they are defined on the GO website. It's an old format, hence fixed columns, no explicit column headings

PR #656, opened 2023-09-22
Adding ability to retrieve axiom annotations on relationships (axiom annotations)

### LINKED ISSUE #654: Add ability to easily retrieve axiom annotations (metadata) on relationships
The `relationships` method yields simple triples, either add a new method to retrieve these with axiom annotations (metadata) -- or add a general axiom_annotations method.

Note that many implementations are already able to do this, the ability is just not yet surfaced.

cmungall:
Fixes #654

Currently only implemented for sql adapter

Example usage:

```python
        >>> from oaklib import get_adapter
        >>> adapter = get_adapter("sqlite:obo:mondo")
        >>> rels = list(adapter.relationships(["MONDO:0009831"]))
        >>> for rel, metadatas in adapter.relationships_metadata(rels):
        ...     for p, v in metadatas:
        ...         print(rel, p, v)
        <BLANKLINE>
        ...
        ('MONDO:0009831', 'rdfs:subClassOf', 'MONDO:0002516') oio:source NCIT:C9005
        ...
```

or on the command line:

```
runoak -i sqlite:obo:mondo relationships MONDO:0009831 --include-metadata -p i
```
```

### DIFF
--- a/src/oaklib/implementations/sqldb/sql_implementation.py
+++ b/src/oaklib/implementations/sqldb/sql_implementation.py
@@ -1062,7 +1062,9 @@ class SqlImplementation(
         for row in q:
             yield row.subject, row.predicate, row.object
 
-    def relationships_metadata(self, relationships: Iterable[RELATIONSHIP], **kwargs) -> Iterator[Tuple[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]]]:
+    def relationships_metadata(
+        self, relationships: Iterable[RELATIONSHIP], **kwargs
+    ) -> Iterator[Tuple[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]]]:
         for rel in relationships:
             anns = [(ann.predicate, ann.object) for ann in self._axiom_annotations(*rel)]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/oaklib/implementations/sqldb/sql_implementation.py [hunk 1]
```diff
@@ -1062,6 +1062,13 @@ def _rbox_relationships(
         for row in q:
             yield row.subject, row.predicate, row.object
 
+    def relationships_metadata(
+        self, relationships: Iterable[RELATIONSHIP], **kwargs
+    ) -> Iterator[Tuple[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]]]:
```
joeflack4, line 1067: I think this should be:
`Iterator[Tuple[RELATIONSHIP, List[Tuple[om.Annotation]]]]`

i.e. replace `PRED_CURIE, Any` with `om.Annotation`

joeflack4 (commented):
Using this now very happy about it!
I missed `_axiom_annotations()` when looking through last week, guess I could have used that too (even though it is private).

PR #659, opened 2023-09-27
Optimization: `SqlImplementation.relationships_metadata()`

### LINKED ISSUE #660: Optimization: `SqlImplementation.relationships_metadata()`
## Overview
@cmungall recently added _ability to retrieve axiom annotations on relationships_ (#656), but there is a significant performance bottleneck. I noticed this during my [`mondo-ingest` synchronization pipeline work](https://github.com/monarch-initiative/mondo-ingest/pull/363#discussion_r1336501888). This operation was taking 104 minutes to complete.

## Possible solutions
- #659

joeflack4:
## Updates
Optimization: SqlImplementation.relationships_metadata()
- Add: SqlImplementation._axiom_annotations_all()
- Update: SqlImplementation.relationships_metadata(): Increased performance to O(log n)
- Update: Added docstrings for several related methods

## Addresses
- #660

### DIFF
--- a/src/oaklib/implementations/sqldb/sql_implementation.py
+++ b/src/oaklib/implementations/sqldb/sql_implementation.py
@@ -1065,8 +1065,61 @@ class SqlImplementation(
     def relationships_metadata(
         self, relationships: Iterable[RELATIONSHIP], **kwargs
     ) -> Iterator[Tuple[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]]]:
-        for rel in relationships:
-            anns = [(ann.predicate, ann.object) for ann in self._axiom_annotations(*rel)]
+        """For given relationships, returns axiom annotation metadata"""
+        # Fetch all annotations
+        anns: List[om.Annotation] = self._axiom_annotations_all()
+        # Simplify objects
+        anns: List[Dict[str, str]] = [{
+            'subject': x.subject,
+            'object': x.object,
+            'predicate': x.predicate,
+            'annotation_predicate': x.annotation_predicate,
+            'annotation_value_or_obj': x.annotation_value if x.annotation_value is not None else x.annotation_object,
+        } for x in anns]
+        # Create lookup for annotations based using edges as keys
+        rel_anns: Dict[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]] = {}
+        for x in anns:
+            key = (x['subject'], x['predicate'], x['object'])
+            if key not in rel_anns:
+                rel_anns[key] = []
+            rel_anns[key].append((x['annotation_predicate'], x['annotation_value_or_obj']))
+        # Group annotations by relationship, and yield
+        rel_anns = {rel: rel_anns[rel] if rel in rel_anns else [] for rel in relationships}
+        for rel, anns in rel_anns.items():
+            yield rel, anns
+
+    def relationships_metadata2(
+        self, relationships: Iterable[RELATIONSHIP], **kwargs
+    ) -> Iterator[Tuple[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]]]:
+        """For given relationships, returns axiom annotation metadata"""
+        rels: List[RELATIONSHIP] = [x for x in relationships]
+        # Split based on predicate
+        preds: Set[PRED_CURIE] = {x[1] for x in rels}
+        rel_anns_all = {}
+        for pred in preds:
+            # Filter rels
+            rels_i = [x for x in rels if x[1] == pred]
+            # Fetch annotations for given relationships
+            subs, objs = [x[0] for x in rels_i], [x[2] for x in rels_i]
+            anns: List[om.Annotation] = [x for x in self._axiom_annotations_multi(subs, pred, objs)]
+            anns: List[Dict[str, str]] = [{
+                'subject': row.subject,
+                'object': row.object,
+                'predicate': row.predicate,
+                'annotation_predicate': row.annotation_predicate,
+                'annotation_value_or_obj': row.annotation_value if row.annotation_value is not None else row.annotation_object,
+            } for row in anns]
+            rel_anns: Dict[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]] = {}
+            for x in anns:
+                key = (x['subject'], x['predicate'], x['object'])
+                if key not in rel_anns:
+                    rel_anns[key] = []
+                rel_anns[key].append((x['annotation_predicate'], x['annotation_value_or_obj']))
+            # Group annotations by relationship
+            rel_anns = {rel: rel_anns[rel] if rel in rel_anns else [] for rel in rels_i}
+            rel_anns_all.update(rel_anns)
+        # Yield
+        for rel, anns in rel_anns_all.items():
             yield rel, anns
 
     def node_exists(self, curie: CURIE) -> bool:
@@ -1374,6 +1427,7 @@ class SqlImplementation(
     def _axiom_annotations(
         self, subject: CURIE, predicate: CURIE, object: CURIE = None, value: Any = None
     ) -> List[om.Annotation]:
+        """Get all axiom annotations for a single edge."""
         q = self.session.query(OwlAxiomAnnotation)
         q = q.filter(OwlAxiomAnnotation.subject == subject)
         q = q.filter(OwlAxiomAnnotation.predicate == predicate)
@@ -1396,6 +1450,7 @@ class SqlImplementation(
         objects: List[CURIE] = None,
         values: List[Any] = None,
     ) -> Iterator[OwlAxiomAnnotation]:
+        """Get all axiom annotations from multiple edges sharing the same predicate."""
         q = self.session.query(OwlAxiomAnnotation)
         if subjects:
             q = q.filter(OwlAxiomAnnotation.subject.in_(subjects))
@@ -1406,6 +1461,10 @@ class SqlImplementation(
             q = q.filter(OwlAxiomAnnotation.value.in_(values))
         for row in q:
             yield row
+
+    def _axiom_annotations_all(self) -> List[OwlAxiomAnnotation]:
+        """Get all axiom annotations."""
+        return self.session.query(OwlAxiomAnnotation).all()
 
     def ancestors(
         self,


src/oaklib/implementations/sqldb/sql_implementation.py
```diff
@@ -1065,8 +1065,58 @@ def _rbox_relationships(
     def relationships_metadata(
```
joeflack4, line 1065: ## Recommended implementation
|n_relationships|seconds_taken|
|---------------|-------------|
|500            |21           |
|5000           |23           |
|37129          |27           |

I wasn't sure if you'd like this approach because it is heavier on memory during the function's runtime, but it seems to work well. I added a simple method `._axiom_annotations_all()` which returns all annotations from the DB. Then the rest of the work is done in Python.

src/oaklib/implementations/sqldb/sql_implementation.py
```diff
@@ -1065,8 +1065,58 @@ def _rbox_relationships(
...
+            key = (x['subject'], x['predicate'], x['object'])
+            if key not in rel_anns:
+                rel_anns[key] = []
+            rel_anns[key].append((x['annotation_predicate'], x['annotation_value_or_obj']))
+        # Group annotations by relationship, and yield
+        rel_anns = {rel: rel_anns[rel] if rel in rel_anns else [] for rel in relationships}
+        for rel, anns in rel_anns.items():
+            yield rel, anns
+
+    def relationships_metadata2(
```
joeflack4, line 1091: ## Alternate implementation
|n_relationships                        |seconds_taken|
|---------------------------------------|-------------|
|500                                    |2            |
|5000                                   |53           |
|37129                                  |516          |

This approach splits `relationships` into groups based on predicates. Then it uses existing `. _axiom_annotations_multi()` method to fetch all annotations for each group of relationships.

joeflack4 (commented):
@cmungall This draft is ready for your eyes.
You recently added ability to retrieve axiom annotations on relationships (#656), but there is a significant performance bottleneck. I noticed this during my [`mondo-ingest` synchronization pipeline work](https://github.com/monarch-initiative/mondo-ingest/pull/363#discussion_r1336501888). This operation was taking 104 minutes to complete.

I've sped this up to 30 seconds to 10 minutes, depending on which implementation you prefer. I prefer the faster method.

I also thought of a possible implementation that uses sqlite and/or SqlAlchemy moreso than Python, but Python is easier for me.

I know that before you can merge this, you need at the very least a test case and codestyle adherence. I will meet with @matentzn and see if he wants me to complete this PR. If not, then this is here for you / your team as an example of what could / should be done.

FYI even though there's no test case, I was able to check that the code works as expected in an ad hoc way, by editing/running this code in the local version of `oaklib` in the `site-packages` of `mondo-ingest` and using it in my [synchronization PR](https://github.com/monarch-initiative/mondo-ingest/pull/363).

src/oaklib/implementations/sqldb/sql_implementation.py
```diff
@@ -1065,8 +1065,58 @@ def _rbox_relationships(
...
+        self, relationships: Iterable[RELATIONSHIP], **kwargs
+    ) -> Iterator[Tuple[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]]]:
+        """For given relationships, returns axiom annotation metadata"""
+        # Split based on predicate
+        preds: Set[PRED_CURIE] = {x[1] for x in relationships}
```
cmungall, line 1096: remember the input is an Iterable, so if it's an iterator, not a collection, this will consume it

src/oaklib/implementations/sqldb/sql_implementation.py
```diff
@@ -1065,8 +1065,58 @@ def _rbox_relationships(
     def relationships_metadata(
         self, relationships: Iterable[RELATIONSHIP], **kwargs
     ) -> Iterator[Tuple[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]]]:
-        for rel in relationships:
-            anns = [(ann.predicate, ann.object) for ann in self._axiom_annotations(*rel)]
+        """For given relationships, returns axiom annotation metadata"""
+        # Fetch all annotations
+        anns: List[om.Annotation] = self._axiom_annotations_all()
```
cmungall, line 1070: I think you are making assumptions about the distributions of numbers of these. Consider a version of omop where every axiom has an annotation

src/oaklib/implementations/sqldb/sql_implementation.py
cmungall: what if n_relationships > 1m?

src/oaklib/implementations/sqldb/sql_implementation.py
```diff
@@ -1065,8 +1065,58 @@ def _rbox_relationships(
...
+    def relationships_metadata2(
+        self, relationships: Iterable[RELATIONSHIP], **kwargs
+    ) -> Iterator[Tuple[RELATIONSHIP, List[Tuple[PRED_CURIE, Any]]]]:
+        """For given relationships, returns axiom annotation metadata"""
+        # Split based on predicate
+        preds: Set[PRED_CURIE] = {x[1] for x in relationships}
+        rel_anns_all = {}
+        for pred in preds:
+            # Fetch annotations for given relationships
+            subs, objs = [x[0] for x in relationships], [x[2] for x in relationships]
```
cmungall, line 1100: don't you want to filter this?

cmungall (changes requested):
I appreciate the time taken here

Your 2nd approach is on the right lines.

I'd simplify

How about instead making a method like `_axiom_annotations_multi`  that takes in a list of tuples, and let sqlite do the optimization?

src/oaklib/implementations/sqldb/sql_implementation.py
joeflack4: Nice catch! This is something I didn't need for my use case so I added it after and hadn't tested it.

I added:
```python
            # Filter rels
            rels = [x for x in relationships if x[1] == pred]
```
And changed references in that for loop to use `rels` instead of `relationships`

src/oaklib/implementations/sqldb/sql_implementation.py
joeflack4: If `n_relationships` > 1m, then this further increases memory burden and that could be too significant for some machines. The higher the `n` is though, the better this implementation performs compared to the "[alternative implementation](https://github.com/INCATools/ontology-access-kit/pull/659#discussion_r1337932004)", although there are other possible implementations, e.g. [leaning more on sqlite](https://github.com/INCATools/ontology-access-kit/pull/659#pullrequestreview-1645528334).

src/oaklib/implementations/sqldb/sql_implementation.py
joeflack4: Good catch! I added a line above adding `rels` and using that variable instead for the rest of the function.
```py
rels: List[RELATIONSHIP] = [x for x in relationships]
```

src/oaklib/implementations/sqldb/sql_implementation.py
joeflack4: By distribution, do you rather mean, "the potentially high number of annotations (returned)", as you query in [this comment](https://github.com/INCATools/ontology-access-kit/pull/659#discussion_r1337998760)?:
> what if n_relationships > 1m?

If that's what you mean, let's continue discussion in that thread and mark this one resolved. If you mean something else though, can you clarify?

src/oaklib/implementations/sqldb/sql_implementation.py
joeflack4: Just in interest of keeping comments in threads / resolvable, replying to your [summary comment](https://github.com/INCATools/ontology-access-kit/pull/659#pullrequestreview-1645528334) here: 

> Your 2nd approach is on the right lines.

I thought you might feel that way. It's currently much less performant than this one, but you have an idea to speed that up:

> I'd simplify; How about instead making a method like `_axiom_annotations_multi` that takes in a list of tuples, and let sqlite do the optimization?

I was thinking of that too, but it might take me significantly more time due to my lack of familiarity. I might not have that time, but I'll ask Nico; maybe I do. If I don't, I was wondering if you wouldn't mind finishing up what I've started?

I'm not sure we need two variations of `._axiom_annotations_multi()`. Why don't we increase its performance? Here's my ideas so far:
<details><summary>._axiom_annotations_multi() optimizations</summary>
<p>

### a. Indexing
Are columns `subject`, `object`, and `predicate` indexed? If not, I wonder if adding indexes might be all we need to make this performant.

### b. (Temporary) lookup column
We can basically do the same thing as my first approach, but in SQL rather than Python. I am creating a lookup using `RELATIONSHIP`s as dictionary keys. We might do the same in sqlite, creating either/both columns that concat `sub-pred-obj` and `sub-pred`:

|sub-pred-obj    |sub-obj    |sub  |pred|obj  |MORE_COLS|
|----------------|-----------|-----|----|-----|---------|
|M:123-is_a-M:456|M:123-M:456|M:123|is_a|M:456|...      |
|M:123-is_a-M:456|M:123-M:456|M:123|is_a|M:456|...      |

Will have some duplicate rows due to multiple axioms per edge, but can still make for faster queries.

Alternatively, we could create a new table with only 1 row per edge, and collect all the axioms for that edge as a JSON serialization in a single column. That should be really fast.

</p>
</details>

matentzn:
Nice I didnt realise you were working on this! Cool!

joeflack4:
@matentzn we discussed it a little bit at one of our recent meetings. I thought you had seen the PR. But yeah basically just got a figure out from you. If you think it's worth my time to complete this. It will speed up any of our scripts that use this function by about an hour or two.

I'm not actually sure now if we will even need this method for the synchronization pipeline now that I understand the work better.

Chris has a suggested implementation but I think it involves using SQL alchemy a bit more. However, I think the bulk of the time on this will take a lot of back and forth with me and Chris on the implementation. I don't know if you will like the idea that I have in mind, which would be creating an on-the-fly column , basically an index.

Alternative solution to that is to create additional indexes, but that would require an update to semantic SQL, not oak. And Chris hasn't gotten back to me yet on what he thinks about that.

PR #666, opened 2023-10-04
Improve usage of SSSOM-py

cthoyt:
This PR makes use of the higher-level interfaces to SSSOM with the 0.4.* series of SSSOM-py

## Depends on

- https://github.com/mapping-commons/sssom-py/issues/439 / https://github.com/mapping-commons/sssom-py/pull/440
- https://github.com/mapping-commons/sssom-py/issues/441


src/oaklib/utilities/lexical/lexical_indexer.py
```diff
@@ -310,24 +311,19 @@ def lexical_index_to_sssom(
...
-    doc = MappingSetDocument(prefix_map=meta.prefix_map, mapping_set=mset)
-    msdf = to_mapping_set_dataframe(doc)
-    num_mappings = len(msdf.df.index)
-    if ensure_strict_prefixes:
-        msdf.clean_prefix_map()
-        if len(msdf.df.index) < num_mappings:
-            raise ValueError("Mappings included prefixes that were not in the prefix map")
+    # TODO what's the business logic for combining these three
+    #  different streams of prefix maps?
+    converter = curies.chain(
```
cthoyt, line 316: Need external guidance here - there are three streams of incoming prefix maps. How should they be prioritized?

src/oaklib/utilities/mapping/sssom_utils.py
```diff
@@ -74,14 +73,7 @@ def emit(self, obj: Mapping):
...
-            mapping_set_id="temp", mappings=self.mappings, license=UNSPECIFIED_MAPPING_SET_ID
-        )
-        if self.ontology_interface:
-            prefix_map = self.ontology_interface.prefix_map()
-        else:
-            prefix_map = {}
-        doc = MappingSetDocument(prefix_map=prefix_map, mapping_set=mset)
-        msdf = to_mapping_set_dataframe(doc)
+        converter = self.ontology_interface.converter if self.ontology_interface else None
+        msdf = MappingSetDataFrame.from_mappings(self.mappings, converter=converter)
```
cthoyt, line 77: default metadata is applied inside this function, so it doesn't need to be done directly

pyproject.toml
```diff
@@ -16,7 +16,7 @@ linkml-runtime = ">=1.5.3"
 linkml-renderer = ">=0.3.0"
 networkx = ">=2.7.1"
 sssom-schema = ">=0.11.0"
-sssom = ">=0.3.38"
+sssom = ">=0.4.0"
```
hrshdhgd, line 19: `sssom==0.4.0rc1` has been released. Updating this dependency will get you further in QC tests.

src/oaklib/utilities/lexical/lexical_indexer.py
hrshdhgd: `oi.converter` can be last. the rest seems fine.

pyproject.toml
cthoyt: @hrshdhgd thanks, it looks like kgcl-schema has some conflicting dependencies now. How do we unpin a maximum version of `curies` for that?

hrshdhgd:
`kgcl-schema = ">=0.6.0"` should fix the current error.

pyproject.toml
hrshdhgd: kgcl-schema = ">=0.6.0" should fix it.

src/oaklib/utilities/lexical/lexical_indexer.py
cthoyt: Done in 17dfea14

pyproject.toml
cthoyt: Thank you!

cthoyt:
looking like it's passing except a spurious bug in the pubmed eutils (unrelated)

cmungall:
I think this is the correct sequence

1. Make a separate PR where we pin sssom <0.4.0
2. Make an OAK release
3. Make a sssom 0.4.0 release
4. Modify this PR to be >= 0.4.0 (potentially <0.5.0 if more breaking changes are expected)

I wish python dependencies didn't lead to dependency hell, but here we are

(I just realized I said "dependency hell" on pull number 666, 🤘🤘🤘 😈 😈 😈 )

hrshdhgd:
Bullets 1, 2 and 3 are done. @cthoyt , could you take care of bullet 4 please?

pyproject.toml
hrshdhgd: As per Chris:
> potentially <0.5.0 if more breaking changes are expected

May need to be "^0.4.0"

cthoyt:
@hrshdhgd done

PR #669, opened 2023-10-13
Use passthrough from curies v0.6.6

cthoyt:
This PR uses the new version of `curies` that simplifies the expand and compress functions to be able to pass through the original value if they can't be operated on. This new feature is disabled by default, but so it is explicitly enabled in several places in OAK to demonstrate its usage

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -23,99 +23,99 @@ test = ["pytest", "pytest-cov"]
 
 [[package]]
 name = "aiohttp"
-version = "3.8.5"
+version = "3.8.6"
 description = "Async http client/server framework (asyncio)"
 category = "main"
 optional = true
 python-versions = ">=3.6"
 files = [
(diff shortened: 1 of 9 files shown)


hrshdhgd (approved):
I don't completely understand what `passthrough` does. Could you provide an example? Trusting you this works!

cthoyt:
@hrshdhgd passthrough does exactly what was implemented before in the diff of this PR - it makes it so if the URI can't be compressed (when running `converter.compress()`) or the CURIE can't be expanded (when using `converter.expand()`), that the input value is returned.

```python
import curies
converter = curies.get_obo_converter()

>>> converter.compress("http://purl.obolibrary.org/obo/GO_1234567")
'GO:1234567'

>>> converter.compress("https://example.org/nope")
None

>>> converter.compress("https://example.org/nope", passthrough=True)
'https://example.org/nope'
```

hrshdhgd:
Aah! Thank you for the clarification!

PR #683, opened 2023-11-22
Obo compliance suite

cmungall:
See

- https://github.com/owlcollab/oboformat/issues/146
- https://github.com/geneontology/obographs/issues/106


matentzn (approved):
This is massively fantastic. I would just like to use it asap. Let suse the current ROBOT release as a baseline. 

What would be nice: How can we _easily_ use this for benchmarking arbitrary tools? Like, what would be the instruction to someone that asks "I have this awesome tool that reads OWL and writes obographs, and I want to make sure it does it correctly". Can we somehow generalise the test method in a way that we can benchmark arbitrary tools?

Also, I prefer this to be in the obographs repo if at all easy to move.

cmungall:
Something is up with this PR. I tried cherry picking all the commits onto a new branch

#693

But still doesn't work...

PR #685, opened 2023-12-05
Update dependencies to open up pydantic v2 requirements downstream

glass-ships:
via updates to kgcl


glass-ships:
never mind, i see harshad's PR which has the same effect, but tests are passing

Issue #323, opened 2022-10-19
Check dependencies in pyproject.toml


cmungall:
- some are not direct dependencies (e.g lark).
- some are pinned to restrictively. This prohibits reuse

PR #686, opened 2023-12-05
OAK dependency tuning

hrshdhgd:
- Addresses [comment](https://github.com/INCATools/ontology-access-kit/pull/666#issuecomment-1839673825) on #666
- Addresses #323

- [x] Limited `sssom` to < 0.4.0
- [x] Removed unnecessary `sssom-schema` dependency since it is satisfied by `sssom` itself.
- [x] Removed `lark` dependency (Unsure how it got in there)
- [x] Added `linkml` as a dependency in tox for running tests.
- [x] Removed `semsimian` fixed version dependency.


pyproject.toml
```diff
@@ -37,7 +35,7 @@ pysolr = "^3.9.0"
 eutils = ">=0.6.0"
 requests-cache = "^1.0.1"
 click = "*"
-semsimian = "0.2.1"
+semsimian = ">=0.2.1"
```
cmungall, line 38: In general I favor less pinning but given how tightly coupled these are, I think limiting to `0.2.*` may be sensible? See also Mark's Q in #oak

tox.ini
```diff
@@ -26,6 +26,7 @@ deps =
     coverage
     semsimian
     gilda
+    linkml
```
cmungall, line 29: a bit worried about duplication of dependency metadata but I don't quite understand why these are in tox.ini in the first place

cmungall (approved):

tox.ini
hrshdhgd: This is because if we run `tox -e py` (to run pytest via tox), it throws errors about linkml module not found (in tot's virtual environment). It doesn't affect `poetry` and it's working in general. We could do away with `tox` and be solely `poetry` dependent but we've had those conversations in the past.

pyproject.toml
hrshdhgd: ok, switched to `^0.2.1`. As per Mark's Q, I did read it and it was confusing with the versions he mentioned.

PR #688, opened 2023-12-11
`.entities()` typedef update

joeflack4:
## Updates
- Update: Type definition for entities() method type definitions. Was previously just Iterable[CURIE], but it is also possible for it to return URIs.

### DIFF
--- a/src/oaklib/implementations/funowl/funowl_implementation.py
+++ b/src/oaklib/implementations/funowl/funowl_implementation.py
@@ -1,6 +1,6 @@
 import logging
 from dataclasses import dataclass
-from typing import Any, Iterable, List, Mapping, Optional
+from typing import Any, Iterable, List, Mapping, Optional, Union
 
 import rdflib
 from funowl import (
@@ -27,7 +27,7 @@ from oaklib.interfaces import SearchInterface
 from oaklib.interfaces.basic_ontology_interface import LANGUAGE_TAG
 from oaklib.interfaces.owl_interface import OwlInterface, ReasonerConfiguration
 from oaklib.interfaces.patcher_interface import PatcherInterface
-from oaklib.types import CURIE, PRED_CURIE
+from oaklib.types import CURIE, PRED_CURIE, URI
 
 
 @dataclass
@@ -97,7 +97,7 @@ class FunOwlImplementation(OwlInterface, PatcherInterface, SearchInterface):
             else:
                 raise ValueError(f"Label must be literal, not {label}")
 
-    def entities(self, filter_obsoletes=True, owl_type=None) -> Iterable[CURIE]:
+    def entities(self, filter_obsoletes=True, owl_type=None) -> Iterable[Union[URI, CURIE]]:
         for ax in self._ontology.axioms:
             if isinstance(ax, Declaration):
                 uri = ax.v.full_uri(self.functional_writer.g)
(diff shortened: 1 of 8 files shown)


src/oaklib/implementations/funowl/funowl_implementation.py
```diff
@@ -97,7 +97,7 @@ def label(self, curie: CURIE, lang: Optional[LANGUAGE_TAG] = None) -> str:
             else:
                 raise ValueError(f"Label must be literal, not {label}")
 
-    def entities(self, filter_obsoletes=True, owl_type=None) -> Iterable[CURIE]:
+    def entities(self, filter_obsoletes=True, owl_type=None) -> Iterable[Union[URI, CURIE]]:
```
joeflack4, line 100: ## The update
This is a minor update. I just happened to notice that while I was using this for `SqlImplementation`, I was getting back URIs sometimes. I assume this should always be a possible thing for the other implementations as well.

joeflack4, line 100: ## Alternatives?
LinkML has [`URIorCURIE`](https://github.com/linkml/linkml-runtime/blob/01838a06bd8274bc835be8ecff0a4a44335eeb99/linkml_runtime/utils/metamodelcore.py#L100) but I don't know if it's something that should be used here.

joeflack4, line 100: ## Order of types?
This is a super minor thing, but I am going to guess that if there is a Python convention, maybe it should be `Union[CURIE, URI]` for alphabetical reasons, but I hose `Union[URI, CURIE]` because that's usually the order we speak them in the semantic web world.

hrshdhgd: No The URI and CURIE definition are `oaklib` specific. So the sis correct.

hrshdhgd: Thanks @joeflack4 !

hrshdhgd: Does not matter.

hrshdhgd (commented):
I think it's good. I'll let @cmungall  make the final decision.

cmungall:
Hi @joeflack4 - it's best to discuss what you want to do in an issue first. The intent is that CURIEs are input and output. Yes, I know that when using semsql URIs that have no prefixes get turned into quoted URIs (different from URIs), but this should be fixed elsewhere.

Is this a blocker for you?

joeflack4:
@cmungall Not a blocker. Just thought it was a quick way to contribute.

Agreed it's a good practice to make an issue. I thought this one would be an easy merge, but as if you think it requires discussion I can go ahead and make an issue.

Here are my thoughts. I'll also add these to the issue:
<details><summary>Details</summary>
<p>

I guess you're right, they are "quoted URIs". Here's an example of one I see:
> `'<http://identifiers.org/hgnc/10004>'`

Why don't we add something like QUOTED_URI to `oaklib.types`?: 
```py
CURIE = str
QUOTED_URI = str
URI = str
```

When you say "should be fixed elsewhere", I assume you are referring to them being "quoted" URI strings rather than plain URI strings. I'm definitely down with that. If you mean that URIs should never be returned (you probably don't), I don't think that's possible 100% of the time.

</p>
</details>

PR #691, opened 2023-12-20
VSKit  - description newline replacement

plbremer:
VSKit generates "descriptions" for LinkML from ontologies. Sometimes, those descriptions contain newline characters. These newline characters break downstream LinkML tooling such as the Pydantic generator. 

I believe that it is appropriate to replace those newline characters with spaces at this point in the flow. 

An example of this newline phenomenon is the permissible value IP-seq from EFO:

```
      IP-seq:
        text: IP-seq
        description: "IP-seq is an assay in which immunoprecipitation with high throughput\
          \ sequencing is used to identify the DNA/RNA-associated proteins or protein\
          \ complexes.\nAs an immunoprecipitation it can be ChIP, RIP or another immunoprecipitation\
          \ process."
        meaning: EFO:0005032
        title: IP-seq
```

Which does not look the same as, for example, its rendering on EBI
https://www.ebi.ac.uk/ols4/ontologies/efo/classes?short_form=EFO_0005032

### DIFF
--- a/src/oaklib/utilities/subsets/value_set_expander.py
+++ b/src/oaklib/utilities/subsets/value_set_expander.py
@@ -183,6 +183,8 @@ class ValueSetExpander(BasicOntologyInterface, ABC):
         enum_definition: EnumDefinition = None,
     ) -> PermissibleValue:
         definition = oi.definition(curie)
+        # \n can break some downstream tooling like LinkML's gen-pydantic (v1.6.6)
+        definition = definition.replace('\n',' ')
         label = oi.label(curie)
         pv_formula = enum_definition.pv_formula if enum_definition else None
         if str(pv_formula) == "CURIE":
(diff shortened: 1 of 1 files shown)


src/oaklib/utilities/subsets/value_set_expander.py
```diff
@@ -183,6 +183,8 @@ def _generate_permissible_value(
         enum_definition: EnumDefinition = None,
     ) -> PermissibleValue:
         definition = oi.definition(curie)
+        # \n can break some downstream tooling like LinkML's gen-pydantic (v1.6.6)
```
cmungall, line 186: I think this should be fixed in the tooling, as newlines are valid (if unusual) in textual elements. But we can merge this for now

cmungall (approved):

Issue #687, opened 2023-12-06
Dumpers and loaders: Separate concept of syntax and datamodel


cmungall:
Currently the Dumper.dump() method accepts a `syntax` argument, this is sometimes called `format` in other frameworks, e.g. robot.

The problem here is that this is frequently ambiguous, especially in the context of OAK which is pluralistic and supports multiple ways of modeling and serializing ontologies, e.g. owl mapped to rdf serialized as rdf/xml; skos (natively rdf), serialized to turtle.

This is compounded with loaders, where we might want to use a suffix to guess the underlying model and serialization and choose the appropriate parser. Unlike the owlapi, rdflib requires the format of rdf to be known in advance (and in my experience this is a good thing - there is a lot of confusion caused by the owlapi cycling through multiple parsers and models).

Examples:

- `.owl` clearly means the OWL *data model*, in the OBO universe this is *conventionally* mapped to RDF and serialized as RDF/XML, but outside this universe the serialization is more typically Turtle, and may not be an RDF serialization at all
- `.xml` means OWL/XML as far as the OWLAPI is concerned, but rdflib uses this to mean RDF/XML (which is very different!)
- `.rdf` might typically mean some kind of RDF serialization of OWL, but SKOS is valid for the ontology-like artefacts in OAK and it can also be serialized as .rdf. Same for the extended RDFS-like model used by schema.org. On top of this, again, we don't know if this means rdf/xml, rdf/turtle, n-quads...

On top of this, there are various aliases (e.g ttl vs turtle). Frameworks like pyoxigraphs use mime types to try and enforce some kind of standard but this seems overkill

Proposal:

- Loaders and dumpers take an optional `model` argument
- If absent, this is inferred using `syntax` and sensible defaults
- We encourage (but do not mandate) bipartite file suffixes to reduce ambiguity and facilitate default arguments

The syntax for bipartitle syntaxes would be `.model.syntax`. For example, `.owl.ttl`, `.skos.nt`

There is a potential argument for a tripartite model here, because of owl mapping to rdf, and to reduce the ambiguity of `.owl.xml`. However, this is likely overkill. 

Something like

## Unambiguous OWL syntaxes

- `.owx`
- `.ofn`
- `.omn`

Model optional, if specified, MUST be `owl`

## OWL layered on RDF

- `owl.ttl` (aka `turtle`)
- `owl.nt` (aka `ntriples`)
- `owl.rdfxml` (maps to `xml` syntax in rdflib)

[...]

PR #693, opened 2023-12-21
OBO Compliance suite (replay of 683)

cmungall:
- First pass at oboformat and obographs conformance suite.
- Allow choice of format when exporting OWL. Addung missing features to obo conversion
- lint
- lint
- Lint. Removing prints.
- Added format_utilities.py -- see #687
- adding test failure info
- reconciled dumper changes between branches


cmungall:
I can't figure out what is going on here.

The tests work locally, github actions seems to be running against something different than what is on the branch

matentzn (changes requested):
This PR is a bit too big to carefully review for me, so I will trust that it wont result in any breaking. However, can you add a section to the OAK docs on "How to test a new OBOgraphs release with OAK"?

I can see the tests in test_obo_format.py, but I cannot easily imagine the recipe.

1. New OBO graphs version for testing available
2. Compile to JAR
3. Compile new ROBOT JAR with that test version of OBO graphs
4. Run...
5. If X then all is well and we can make an OBO graphs release and push it into ROBOT

A recipe like this would be very useful.

There are also two files

tests/input/obo_conf.yaml
tests/input/obojson_conf.yaml

that are generated as part of a local run. Should these be committed or gitignored?

cmungall:
Works for me locally:

```
poetry run pytest tests/test_utilities/test_cross_ontology_diff.py
```

But this fails on actions

cmungall:
Closed in favor of #715

PR #704, opened 2024-03-06
Made semsimian an optional dependency

### LINKED ISSUE #703: Missing support for Python 3.12
[Claimed support for Python 3.12](https://pypi.org/project/oaklib/) is missing.

The installation fails due to dependencies that cannot be installed on Python 3.12:

* oaklib [depends on semsimian ^0.2.1](https://github.com/INCATools/ontology-access-kit/blob/15bf85cefc2fe8541b38aabfcf7c65eb46bc1231/pyproject.toml#L38)
* no build of semsimian exists for Python 3.12, currently [builds exist for 3.7 =< Python =< 3.11](https://github.com/monarch-initiative/semsimian/blob/0ae82eb20c169037344de1bcbb5523a1887c95f7/.github/workflows/release.yml).

This issue is [blocking support for Python 3.12 in `schema-automator`](https://github.com/linkml/schema-automator/issues/134).

hrshdhgd:
Fixes #703


cmungall (approved):

cmungall (changes requested):
I think this might break things as-is

https://github.com/INCATools/ontology-access-kit/blob/15bf85cefc2fe8541b38aabfcf7c65eb46bc1231/src/oaklib/implementations/__init__.py#L44-L46

https://github.com/INCATools/ontology-access-kit/blob/15bf85cefc2fe8541b38aabfcf7c65eb46bc1231/src/oaklib/implementations/semsimian/semsimian_implementation.py#L9

hrshdhgd:
> I think this might break things as-is
> 
> https://github.com/INCATools/ontology-access-kit/blob/15bf85cefc2fe8541b38aabfcf7c65eb46bc1231/src/oaklib/implementations/__init__.py#L44-L46
> 
> https://github.com/INCATools/ontology-access-kit/blob/15bf85cefc2fe8541b38aabfcf7c65eb46bc1231/src/oaklib/implementations/semsimian/semsimian_implementation.py#L9

Aah! great catch ...that would have introduced more headaches. I think this is why we had `semsimian` as a required package. Thinking of a smart solution.....

cmungall (changes requested):
I think this is a bit complicated and will raise an error if someone imports semsimian_implementation. In general it should be safe to import any module *in* the current package.

Look at gilda_implementation for inspiration

cmungall (changes requested):
There should be no need for dynamic manipulation of `__all__`

This can be simplified further. When I say look at gilda, I mean make it completely parallel. It's completely analogous, gilda==semsimian. `__init__.py` still imports `GildaImplementation`. It's completely safe as gilda is only ever imported within a method that is explicitly executed by a client

hrshdhgd:
Done

src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -22,6 +20,8 @@
 from oaklib.interfaces.semsim_interface import SemanticSimilarityInterface
 from oaklib.types import CURIE, PRED_CURIE
 
+Semsimian: Optional["Semsimian"] = None  # type: ignore
```
cmungall, line 23: sorry... still pretty confusing. So this is a global variable? With a leading uppercase that makes it sound like a class?

hrshdhgd: It is a class but it is instantiated as a `None`. The class only exists once it is imported from the `semsimian` package. I know it is super confusing. I still prefer my previous solution but this one works too. Mimicking `Gilda` comes with complications like this because we are caching the Semsimian object but Gilda does not cache anything. So something that works for Gilda implementation doesn't necessaarily seem to work for Semsimian implementation.

cmungall:
You don't need to worry about caching. In fact this could be problematic.

When you call get_adapter() it creates an SSI instance and this can hold on
to the SS object

cmungall (approved):

PR #717, opened 2024-03-14
Update semsimian dependency version

hrshdhgd:
- [x] Update semsimian dependency to latest version
- [x] Converted a bioportal implementation test to a `mock` API call since the test failed/is outdated with regards to data returned by the API call made.


cmungall (commented):
Did you mean to update the pyproject.toml too?

hrshdhgd:
Since it's an extra the toml file has it without a version number

hrshdhgd:
it was already `>=0.1.0`, so should work without a change. However, I did make the change just to be sure.

cmungall (approved):

PR #719, opened 2024-03-18
More efficient simple obo diffs

cmungall:
- Adding a more efficient diff implementation for simpleobo
    - technically this is a violation of the DRY principle but one of the goals of the interface/impl split is to allow implementations to do custom implementations that may be faster
 - Undoing some of https://github.com/INCATools/ontology-access-kit/pull/605

### DIFF
--- a/src/oaklib/implementations/simpleobo/simple_obo_implementation.py
+++ b/src/oaklib/implementations/simpleobo/simple_obo_implementation.py
@@ -61,7 +61,7 @@ from oaklib.datamodels.vocabulary import (
     SKOS_CLOSE_MATCH,
     SUBPROPERTY_OF,
     TERM_REPLACED_BY,
-    TERMS_MERGED,
+    TERMS_MERGED, SCOPE_TO_SYNONYM_PRED_MAP,
 )
 from oaklib.implementations.simpleobo.simple_obo_parser import (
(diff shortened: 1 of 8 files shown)


src/oaklib/implementations/simpleobo/simple_obo_implementation.py
```diff
@@ -761,9 +769,161 @@ def logical_definitions(
     # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
     # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-    # Implements: PatcherInterface
+    # Implements: DifferInterface
     # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+    def diff(
```
hrshdhgd, line 775: The `diff` in the `diff_interface.py` already works fine for obo files. I ask just out of curiosity, how is this implementation different?

hrshdhgd (approved):

src/oaklib/implementations/simpleobo/simple_obo_implementation.py
cmungall: it's faster!

PR #720, opened 2024-03-19
Handle `RemoveTextDefinition`

hrshdhgd:
This change handling for markdown format did not exist before. Also added old value for definition in the class.

### DIFF
--- a/src/oaklib/implementations/simpleobo/simple_obo_implementation.py
+++ b/src/oaklib/implementations/simpleobo/simple_obo_implementation.py
@@ -861,7 +861,7 @@ class SimpleOboImplementation(
                         id=_id(), about_node=t1id, new_value=td2, old_value=td1
                     )
                 elif vals1:
-                    yield kgcl.RemoveTextDefinition(id=_id(), about_node=t1id, old_value=vals1.pop())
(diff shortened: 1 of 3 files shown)


src/oaklib/implementations/simpleobo/simple_obo_implementation.py
```diff
@@ -861,7 +861,7 @@ def _tv_dict(stanza: Stanza) -> Dict[str, List[str]]:
                         id=_id(), about_node=t1id, new_value=td2, old_value=td1
                     )
                 elif vals1:
-                    yield kgcl.RemoveTextDefinition(id=_id(), about_node=t1id)
+                    yield kgcl.RemoveTextDefinition(id=_id(), about_node=t1id, old_value=vals1.pop())
```
cmungall, line 864: don't mutate, lookup instead! otherwise all good, thx!

cmungall (changes requested):

src/oaklib/implementations/simpleobo/simple_obo_implementation.py
hrshdhgd: fixed. It's already there in `td1`

cmungall (approved):

PR #721, opened 2024-03-19
Added mock tests

hrshdhgd:
`TestNCBIGene` => `test_query` consistently fails at first go until it passes after a workflow rerun. Converted into a mock test to address this


tests/test_implementations/test_ncbigene.py
```diff
@@ -8,26 +9,31 @@
 )
 
 from tests import CYTOPLASM, INPUT_DIR
+from tests.test_implementations.test_mock_objects import NCBI_ASSOCIATIONS
```
cmungall, line 12: rename to test_mock_ncbi_objects

cmungall: or input/mock_ncbi_objects

cmungall (approved):

PR #723, opened 2024-03-21
Fixed test

hrshdhgd:
- [x] Need to unskip this test pending a new release of kgcl-schema


tests/test_implementations/test_robot_template.py
```diff
@@ -66,12 +65,12 @@ def test_patcher(adapter):
         print("NEW:", change_obj.new_value)
         adapter.apply_patch(change_obj)
     # adapter.set_label(BRAIN_SPECIMEN, "FOO")
-    assert adapter.label(BRAIN_SPECIMEN) == "FOO"
+    assert adapter.label(BRAIN_SPECIMEN) == "brain sample"
```
cmungall, line 68: umm no, the whole point is we want to test the applied change

cmungall (changes requested):

tests/test_implementations/test_robot_template.py
```diff
@@ -65,7 +64,7 @@ def test_patcher(adapter):
         assert isinstance(change_obj, kgcl.NodeRename)
         print("NEW:", change_obj.new_value)
         adapter.apply_patch(change_obj)
-    # adapter.set_label(BRAIN_SPECIMEN, "FOO")
+    adapter.set_label(BRAIN_SPECIMEN, "FOO")
```
cmungall, line 67: the point of the test is to test apply_patch. The commented out line was left over from previous experimentation. It should just be removed. We already have a test for set_label

cmungall (changes requested):

cmungall (approved):

PR #741, opened 2024-04-25
Enriched mapping documentation

cmungall:
- added a Commands/Mappings notebook
- extended the OAK mappings guide to talk about some difficulties around directionality
- clarified how weights work in lexmatch rules
- removed duplicate synonymizer section in rules data model (it now has its own data model)

### DIFF
--- a/docs/guide/mappings.rst
+++ b/docs/guide/mappings.rst
@@ -26,11 +26,13 @@ mappings are represented as simple lists of mapped CURIEs, without any metadata.
 Mappings in OAK
 ---------------
 
-To see all mappings for a concept or concept, you can use the mappings command:
+To see all mappings for a concept or concept, you can use the mappings command.
+
+See the `Notebook example on mappings <https://github.com/INCATools/ontology-access-kit/blob/main/notebooks/Commands/Mappings.ipynb>`_.
 
 .. code-block:: bash
 
-    $ runoak -i sqlite:obo:uberon mappings brain
+    $ runoak -i sqlite:obo:uberon mappings UBERON:0000955
     ...
     ---
     subject_id: UBERON:0000955
@@ -51,7 +53,7 @@ To get this as SSSOM TSV:
 
 .. code-block:: bash
 
-    $ runoak -i sqlite:obo:uberon mappings brain --output-type sssom
+    $ runoak -i sqlite:obo:uberon mappings UBERON:0000955 --output-type sssom
 
 .. csv-table:: uberon sssom mappings
     :header: subject_id,subject_label,predicate_id,object_id,mapping_justification,subject_source,object_source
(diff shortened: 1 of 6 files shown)


gouttegd (commented):
Looks good to me, apart from a confusing statement about the directionality of mappings (see embedded comment).

notebooks/Commands/Mappings.ipynb
```diff
@@ -0,0 +1,603 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "32a2e447",
+   "metadata": {},
+   "source": [
+    "# OAK mappings command\n",
+    "\n",
+    "This notebook is intended as a supplement to the [main OAK CLI docs](https://incatools.github.io/ontology-access-kit/cli.html).\n",
+    "\n",
+    "This notebook provides examples for the `mappiungs` command, which can be used to lookup mappings that are bundled with ontologies.\n",
```
gouttegd, line 12:
```suggestion
    "This notebook provides examples for the `mappings` command, which can be used to lookup mappings that are bundled with ontologies.\n",
```

gouttegd:
Is the `mappings` command supposed to only be used with the SQLite adapter (as in the examples provided in the new notebook)?

Because when I try to use it directly on a OBO file, it gives a completely different output:

```sh
$ runoak -i uberon.obo mappings UBERON:0000995
This ontology interface contains 267 prefixes:
  AAO -> http://purl.obolibrary.org/obo/AAO_
  ADO -> http://purl.obolibrary.org/obo/ADO_
  ADW -> http://purl.obolibrary.org/obo/ADW_
  AEO -> http://purl.obolibrary.org/obo/AEO_
[…]
```

It goes to list all the prefixes, but never actually lists any mapping.

Whereas using it through the SQLite adapter gives the same results as in the doc:

```sh
$ runoak -i sqlite:obo:uberon mappings UBERON:0000995
subject_id: UBERON:0000995
predicate_id: oio:hasDbXref
object_id: BTO:0001424
mapping_justification: semapv:UnspecifiedMatching
subject_label: uterus
subject_source: UBERON
object_source: BTO
[…]
```

Is that a bug or the expected behaviour? If the latter, this warrants a mention in the doc.

docs/guide/mappings.rst
```diff
@@ -84,12 +86,91 @@ In Python
     UBERON:0000955 EFO:0000302 oio:hasDbXref
     ...
 
+Directionality of mappings
+--------------------------
+
+While mappings are often thought of as unidirectional, this is not quite true in practice:
```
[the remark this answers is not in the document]
cmungall: I do indeed mean the opposite of what I said!

cmungall:
Hmm, on my phone now so can't check now but I believe that behavior is when pronto can't parse the obo

docs/guide/mappings.rst
```diff
@@ -84,12 +86,91 @@ In Python
...
 
+Directionality of mappings
+--------------------------
+
+While mappings are often thought of as bidirectional, this is not quite true in practice:
+
+- Some mapping predicates such as ``skos:broadMatch`` and ``skos:narrowMatch`` are inherently directed, with the meaining inverted in the opposite direction
+- Many mappings are only be asserted in one ontology, and may be "invisible" when queried from the mapped ontology
+
+At the time of writing, most ontologies bundle mappings that are uncommitted `oio:hasDbXref`.
```
twhetzel, line 97: Is the prefix `oio` mentioned earlier or 110% known by readers what this is?

docs/guide/mappings.rst
```diff
@@ -84,12 +86,91 @@ In Python
     UBERON:0000955 EFO:0000302 oio:hasDbXref
     ...
 
+Directionality of mappings
+--------------------------
+
+While mappings are often thought of as bidirectional, this is not quite true in practice:
+
+- Some mapping predicates such as ``skos:broadMatch`` and ``skos:narrowMatch`` are inherently directed, with the meaining inverted in the opposite direction
```
twhetzel, line 94: directed --> directional?

this `with the meaining inverted in the opposite direction` may not be very clear and also just noticed typo in `meaining`

docs/guide/mappings.rst
```diff
@@ -84,12 +86,91 @@ In Python
...
+- Many mappings are only be asserted in one ontology, and may be "invisible" when queried from the mapped ontology
```
twhetzel, line 95: Might want to expand on why mappings may be "invisible" from an ontology, e.g. each ontology must agree to the mapping and include the mapping and actively assert the mapping to the term in the other ontology for it to be visible from either ontology.

src/oaklib/datamodels/mapping_rules_datamodel.yaml
```diff
@@ -74,61 +80,43 @@ classes:
     attributes:
       subject_source_one_of:
         multivalued: true
+        description: The source of the subject to be matched.
+          Multiple values can be provided, it must match at least one.
```
twhetzel, line 84: reading only from the gh diff, an example value looks like it would be useful to add

PR #747, opened 2024-05-01
Update obo compliance suite using latest robot

cmungall:
Using robot prerelease from https://github.com/owlcs/owlapi/pull/1102


cmungall:
The large diff is explained by https://github.com/owlcs/owlapi/issues/1134

Let's just accept this for now, so we are in sync with the latest owlapi

PR #755, opened 2024-05-08
Update to support new semsimian IC map functionality

caufieldjh:


src/oaklib/implementations/semsimian/semsimian_implementation.py
```diff
@@ -110,6 +108,110 @@ def _get_semsimian_object(
 
         return self.semsimian_object_cache[predicates]
 
+    def get_information_content(
```
cmungall, line 111: keep delegating for now

cmungall (approved):

PR #759, opened 2024-05-13
Pass custom IC map to Semsimian through CLI

caufieldjh:


caufieldjh:
See also https://github.com/monarch-initiative/semsimian/issues/127 - if Semsimian encounters an error in parsing the provided IC map, it will raise an error but continue.

caufieldjh:
Something like the following should now work:
```
runoak -vvv -i semsimian:sqlite:obo:phenio similarity --no-autolabel -p i -O csv -o HP_vs_HP_semsimian_test.tsv --min-ancestor-information-content 5 --information-content-file test_ic.tsv HP:0000118 @ HP:0000119 
```
with `test_ic_.tsv` containing just two columns, like so:
```tsv
HP:0000118  5.5
HP:0000119  6.5
```

src/oaklib/cli.py
```diff
@@ -3038,7 +3042,10 @@ def termset_similarity(
     if not isinstance(impl, SemanticSimilarityInterface):
         raise NotImplementedError(f"Cannot execute this using {impl} of type {type(impl)}")
     if information_content_file:
-        impl.cached_information_content_map = load_information_content_map(information_content_file)
+        if isinstance(impl, SemSimianImplementation):
+            impl.custom_ic_map_path = information_content_file
+        else:
+            impl.cached_information_content_map = load_information_content_map(information_content_file)
```
cmungall, line 3048: one possibility in future is to relieve client of the need for out of band knowledge about impl details. The generic SemSim interface could have a load_ic_map method, with the generic impl being to directly load, and the semsimian impl passing the path through.

tests/test_implementations/test_semsimian_implementation.py
```diff
@@ -134,6 +136,29 @@ def test_all_by_all_pairwise_similarity(self):
...
+            raise AssertionError("SemanticSimilarityInterface not implemented")
+        entities = [VACUOLE, ENDOMEMBRANE_SYSTEM]
+
+        for s in entities:
+            for o in entities:
+                for preds in [self.predicates]:
+                    sim = adapter.pairwise_similarity(s, o, predicates=preds)
+                    if sim is not None:
+                        if s == VACUOLE and o == VACUOLE:
+                            self.assertGreaterEqual(sim.ancestor_information_content, 5.5)
```
cmungall, line 154: Why not test for near-equality?

caufieldjh: This could actually just be equality as it's reading the score directly from the provided IC map.

PR #764, opened 2024-05-15
More efficient association-based IC calculation.

### LINKED ISSUE #763: Calculating ICs with an annotation corpus as background is slow
the following is slow

```
runoak -g ~/tmp/phenotype.hpoa -G hpoa -i sqlite:obo:phenio information-content -p i --use-associations .all
```

the reasons for this are as follows

- phenio is a sqlite database
- the associations come in separately from a file, so we can't do a join over both these together, utilizing sqlite join optimizations
- the default fallback is very naive, iterating one term at a time

As an immediate fix, the associations could be loaded into the same database as phenio. This is not ideal as this is an awkward step and it can be beneficial to keep these separate

As a medium term fix we should see if there is a way to attach the associations separately (this is easy with duckdb)

However, the immediate short term fix is to use a less naive non-iterative approach to calculating IC

cmungall:
Fixes #763

### DIFF
--- a/src/oaklib/implementations/semsimian/semsimian_implementation.py
+++ b/src/oaklib/implementations/semsimian/semsimian_implementation.py
@@ -14,6 +14,7 @@ from oaklib.datamodels.similarity import (
 )
 from oaklib.datamodels.vocabulary import OWL_THING
 from oaklib.implementations.sqldb.sql_implementation import SqlImplementation
+from oaklib.interfaces.association_provider_interface import AssociationProviderInterface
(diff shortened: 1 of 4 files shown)


justaddcoffee:
Now completes in < 1m! Thanks @cmungall 
```
time runoak -g hpoa.tsv -G hpoa -i sqlite:obo:phenio information-content -p i --use-associations .all
[snip]
real    0m37.716s
user    0m34.187s
sys     0m3.631s
```

justaddcoffee (approved):

Issue #784, opened 2024-07-23
vskit expander should support PVs in the MIxS label [id] style


turbomam:
Ideally `pv_formlua` would accept `MIXS` as an option

I will illustrate with code that only does MIXS style expansion

files added to repo root are not intended for merging... would be replaced with tests

turbomam:
branch `mixs-style-expander`

turbomam:
```shell
vskit expand --schema mixs_inspired_schema.yaml --output mixs_inspired_schema_expanded.yaml 
linkml-validate --schema mixs_inspired_schema_expanded.yaml --target-class Sample mixs_inspired_data.yaml 
```

turbomam:
```shell
vskit expand \
    --schema mixs_inspired_schema.yaml \
    --output mixs_inspired_schema_expanded.yaml \
    --pv-syntax '{label} [{id}]'
```

PR #785, opened 2024-07-23
structured_format.pattern-style permissible value creation eg for MIxS

turbomam:
- https://github.com/INCATools/ontology-access-kit/issues/784

I destructively edited `src/oaklib/utilities/subsets/value_set_expander.py`. I think the real solution will require adding a new permissible value for the `pv_formula` slot in the metamodel

my idea can be interactively illustrated with

```shell
vskit expand --schema mixs_inspired_schema.yaml --output mixs_inspired_schema_expanded.yaml 
linkml-validate --schema mixs_inspired_schema_expanded.yaml --target-class Sample mixs_inspired_data.yaml 
```


turbomam:
in support of
- https://github.com/microbiomedata/issues/issues/468

cc @mslarae13

cmungall:
This should be done by passing a syntax string like `“{label} [{id}]”`

- Longer term: add linkmlml metamodel element for pv_syntax (e.g. “{label} {id}”
- Interim: Do this as a command line option on vskit E.g. –pv-syntax  “{label} {id}”

turbomam:
will remove schema and data file from root now

turbomam:
writing a test

turbomam:
@cmungall I'm going to call it a day now. I added a test that instantiates test data against a schema after expansion. I included the schema as a string in the test and generated my target class for the instantiation with `PythonGenerator.serialize` followed by the builtin `compile` in `exec` mode, which ruff doesn't like, and I don't blame it.

Isn't there a pure LinkML way to do that? I can't remember it.

cmungall (approved):

PR #791, opened 2024-08-15
Implementing NodeDeletion which was earlier a comment.

hrshdhgd:
I think when this code was written `NodeDeletion` was not available in`kgcl_schema`. Now it surely is and hence implementing it and deleting the comment.

### DIFF
--- a/src/oaklib/implementations/simpleobo/simple_obo_implementation.py
+++ b/src/oaklib/implementations/simpleobo/simple_obo_implementation.py
@@ -912,9 +912,8 @@ class SimpleOboImplementation(
                         id=_id(), about_node=t1id, new_value=vals2list[0], old_value=None
                     )
                 else:
-                    #! KGCL may have NameDeletion in the future.
-                    yield kgcl.NodeRename(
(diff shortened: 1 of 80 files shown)


src/oaklib/implementations/simpleobo/simple_obo_implementation.py
```diff
@@ -912,9 +912,8 @@ def _tv_dict(stanza: Stanza) -> Dict[str, List[str]]:
                         id=_id(), about_node=t1id, new_value=vals2list[0], old_value=None
                     )
                 else:
-                    #! KGCL may have NameDeletion in the future.
-                    yield kgcl.NodeRename(
-                        id=_id(), about_node=t2id, old_value=vals2list[0], new_value=None
+                    yield kgcl.NodeDeletion(
+                        id=_id(), about_node=t1id, old_value=vals1list[0], new_value=None
```
hrshdhgd, line 916: @cmungall : This is the only real change in the PR. I ran `tox -e format` that changed al other files. 

Reason: GH Actions were failing and complaining about imports not being sorted.

PR #799, opened 2024-08-18
Add cache management features

### LINKED ISSUE #792: Allow user to explicitly manage pystow cache (for downloaded sqlite dbs)
Copy from Slack:

@turbomam wrote:

I am writing about how great OAK is for a NMDC value set task. But I just realized that this
local/biome-info.txt:
	$(RUN) runoak --input sqlite:obo:envo info  .desc//p=i ENVO:00000428 > $@
does not retrieve [forest biome](https://www.ebi.ac.uk/ols4/ontologies/envo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FENVO_01000174?lang=en) or any of its children

UPDATE_ My cached ~/.data/oaklib/envo.db was from March 9th 2023! (edited) 

Chris Mungall
  [Today at 11:35 AM](https://obo-communitygroup.slack.com/archives/C03D93DEALA/p1723746910572059?thread_ts=1723746222.431679&cid=C03D93DEALA)
Answered separately but the issue here is that it’s easy for cached versions to become stale. There is some discussion here: https://github.com/cthoyt/pystow/issues/54

I think OAK should more actively manage the cache for you but open to ideas about how this should be done

gouttegd:
This PR implements roughly what was proposed in [this comment](https://github.com/INCATools/ontology-access-kit/issues/792#issuecomment-2292350550).

From a user’s perspective:

By default, whenever an attempt is made to access a SQLite DB through the `sqlite:obo:...` descriptor, and the requested DB is already present in the Pystow cache, OAK will check whether the SQLite file is older than 7 days, and if it is, will forcefully re-download that file again.

A new command-line global option `--caching` is available to alter that default behaviour. That option allows to:

* Change the duration after which a cached DB is considered “stale” and in need of refreshing; for example, with `--caching=3w`, a cached DB will be refreshed upon access if is older than 3 weeks (the general syntax is `ND`, where _N_ is a number and `D` can be `s`, `d`, `w`, `m`, or `y` to indicate that _N_ is a number of seconds, days, weeks, months, or years, respectively).
* Prevent OAK from refreshing a cached DB regardless of how old it is, with `--caching=no-refresh`.
* Force OAK to refresh a cached DB regardless of how old it is, with `--caching=refresh`.
* Force OAK to completely clear the cache (which will obviously trigger a refresh), with `--caching=clear` or `--caching=reset`.

Implementation details:

Most of the new code is in the new module `oaklib.utilities.caching` and split in two classes:

* `CachePolicy` represents the logic to determine whether a given file is in need of refreshing.
* `FileCache` represents the file cache itself and is the main interface that the rest of OAK should interact with whenever they need to access the cache (instead of using Pystow directly). It is merely a thin layer on top of Pystow, that tries¹ to present the same interface (same methods) than a Pystow module, so that it can be used as a drop-in replacement.

(¹ I say “tries to” because it only implements the Pystow methods that were actually used somewhere in OAK code, and only with the parameters used in those case.)

A new `FILE_CACHE` object (an instance of `FileCache`) is added to the globals in `oaklib.constants` (same place containing the `PYSTOW_MODULE`), where it can be used by any part of OAK that needs to interact with the cache. The `llm_implementation` and `sqldb_implementation` modules are amended to use that new global instead of `PYSTOW_MODULE`.

Finally, in the main entry point module:

* the `--caching` option is added;
* the `cache-ls` and `cache-clear` commands are re-written to use methods from the `FileCache` class.

closes #792

### DIFF
--- a/docs/intro/tutorial07.rst
+++ b/docs/intro/tutorial07.rst
@@ -64,6 +64,25 @@ This will download the pato.db sqlite file once, and cache it.
 
 PyStow is used to cache the file, and the default location is ``~/.data/oaklib``.
 
+By default, a cached SQLite file will be automatically refreshed (downloaded
+again) if it is older than 7 days. That behavior can be controlled with the
+global ``--caching`` option. For example, to force OAK to always download the
+file regardless of its age:
+
+.. code-block::
+
+    runoak --caching=refresh -i sqlite:obo:pato search t~shape
+
+Other possible values for the ``--caching`` option include:
+
+- ``no-refresh`` to prevent OAK from re-downloading the file even it is older
+  than 7 days;
+- ``Xd`` to refresh a cached file older than _X_ days;
+- ``Xw`` to refresh a cached file older than _X_ weeks.
+
+You may also use the ``cache-clear`` command to force clearing any cached
+SQLite file at anytime.
+
 Building your own SQLite files
 -------------------
 
--- a/docs/intro/tutorial07.rst
+++ b/docs/intro/tutorial07.rst
@@ -65,23 +65,8 @@ This will download the pato.db sqlite file once, and cache it.
 PyStow is used to cache the file, and the default location is ``~/.data/oaklib``.
 
 By default, a cached SQLite file will be automatically refreshed (downloaded
-again) if it is older than 7 days. That behavior can be controlled with the
-global ``--caching`` option. For example, to force OAK to always download the
-file regardless of its age:
-
-.. code-block::
-
-    runoak --caching=refresh -i sqlite:obo:pato search t~shape
-
-Other possible values for the ``--caching`` option include:
-
-- ``no-refresh`` to prevent OAK from re-downloading the file even it is older
-  than 7 days;
-- ``Xd`` to refresh a cached file older than _X_ days;
-- ``Xw`` to refresh a cached file older than _X_ weeks.
-
-You may also use the ``cache-clear`` command to force clearing any cached
-SQLite file at anytime.
+again) if it is older than 7 days. For details on how to alter the behavior of
+the cache, see the :ref:`Cache Control` section in the CLI documentation.
 
 Building your own SQLite files
 -------------------
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -91,6 +91,55 @@ and tracing upwards through is_a and part_of relationships:
 
     uberon viz -p i,p hand foot
 
+Cache Control
+-------------
+
+OAK may download data from remote sources as part of its normal operations. For
+example, using the :code:`sqlite:obo:...` input selector will cause OAK to
+fetch the requested Semantic-SQL database from a centralised repository.
+Whenever that happens, the downloaded data will be cached in a local directory
+so that subsequent commands using the same input selector do not have to
+download the file again.
+
+By default, OAK will refresh (download again) a previously downloaded file if
+it was last downloaded more than 7 days ago.
+
+The global option :code:`--caching` gives the user some control on how the
+cache works.
+
+To change the default cache expiry lifetime of 7 days, the :code:`--caching`
+option accepts a value of the form :code:`ND`, where *N* is a positive integer
+and *D* can be either :code:`s`, :code:`d`, :code:`w`, :code:`m`, or :code:`y`
+to indicate that *N* is a number of seconds, days, weeks, months, or years,
+respectively. If the *D* part is omitted, it defaults to :code:`d`.
+
+For example, :code:`--caching=3w` instructs OAK to refresh a cached file if it
+was last refreshed 21 days ago.
+
+The :code:`--caching` option also accepts the following special values:
+
+- :code:`refresh` to force OAK to always refresh a file regardless of its age;
+- :code:`no-refresh` to do the opposite, that is, preventing OAK from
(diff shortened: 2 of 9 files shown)


docs/intro/tutorial07.rst
```diff
@@ -64,6 +64,25 @@ This will download the pato.db sqlite file once, and cache it.
...
+
+Other possible values for the ``--caching`` option include:
+
+- ``no-refresh`` to prevent OAK from re-downloading the file even it is older
+  than 7 days;
+- ``Xd`` to refresh a cached file older than _X_ days;
+- ``Xw`` to refresh a cached file older than _X_ weeks.
+
+You may also use the ``cache-clear`` command to force clearing any cached
+SQLite file at anytime.
```
matentzn, line 84: So `runoak --caching= clear -i sqlite:obo:pato search t~shape` will clear the entire cache, not just delete the pato cached file?

matentzn:
This is so awesome, thank you!

docs/intro/tutorial07.rst
gouttegd: Yes. That’s the difference between `--caching=clear` (or `--caching=reset`, the two are equivalent) and `--caching=refresh`.

```sh
$ runoak --caching=refresh -i sqlite:obo:path search t~shape
```

will forcefully re-download `pato.db` regardless of its age, but will not touch any other cached file.

```sh
$ runoak --caching=clear -i sqlite:obo:path search t~shape
```

will delete _all_ cached database files regardless of their age. Then the cache code will find that `pato.db` is not in the cache (it just deleted that file itself!), and so will download it again.

So after a command in which you used `--caching=clear`, any subsequent command that uses another database will need to download it again, even if that subsequent command does not explicitly request a refresh.

gouttegd: Note that the `--caching=clear` _option_ is not the same thing as the `cache-clear` _command_.

The `cache-clear` command will forcefully delete any cached database that is older than 100 days (or the number of days specified by the `--days-old` option), regardless of any caching policy.

So, the equivalent of the `--caching=clear` option would be `clear-cache --days-old=0` (remove any file “older than 0 day“), except that `--caching=clear` option would only take effect _when a command actually tries to use the cache_ (e.g. a command involving a `sqlite:obo:...` selector), whereas `clear-cache --days-old=0` would take effect immediately and unconditionally.

(IMO there would no longer be a real need for `clear-cache` if this PR is applied, as that command was merely a crutch for the lack of proper cache management. But that command existed before, so it should be kept.)

matentzn: Sounds great! In this case, I would copy/paste your comments in the tutorial docs so that others understand these distinctions as well!

gouttegd: I don’t think a _tutorial_ should be cluttered with exhaustive documentation about all the precise aspects of a given feature.

I put the bit of doc in the `tutorial07.rst` document because unless I missed something that is one of the the two only places in the entire doc where the fact that SQLite DB files can be automatically downloaded and cached somewhere.

The other place is the [Getting Started](https://incatools.github.io/ontology-access-kit/intro/tutorial01.html#query-the-obo-library) document, which says:

> The base command takes an --input (or just -i) option that specifies the input selector. We will return to the full syntax later, but one pattern that is useful to know is sqlite:obo:<ontology-id>, which uses the SQL Database Adapter to fetch and use the SQLite version of an ontology.
> […]
> The first time you run this there will be a lag as the file is downloaded, but after that it will be cached.

And a “Getting Started” document is, I believe, no more an appropriate place for exhaustive documentation than a “tutorial”.

Exhaustive explanation about how caching works and is controlled would rather belong to _reference documentation_, but the mere concept of caching is currently completely absent of OAK’s reference docs, so I had no place where I could add anything.

For example, the reference documentation about the [selectors](https://incatools.github.io/ontology-access-kit/packages/selectors.html) merely has this to example to show:

> `sqlite:obo:go` - pre-made SQLite from Semantic SQL registry

And the reference documentation for the [SQL Database Adapter](https://incatools.github.io/ontology-access-kit/packages/implementations/sqldb.html) does not say anything at all about automatic download and caching.

gouttegd: I’ve added a new section in the CLI reference documentation to fully explain how the `--caching` option works.

matentzn: This is perfect, thanks @gouttegd!

caufieldjh:
Thanks @gouttegd , this is going to help to avoid a lot of confusion in the future.

gouttegd:
PR updated with the feature discussed [in this comment](https://github.com/INCATools/ontology-access-kit/issues/792#issuecomment-2296908819).

Briefly, in addition to the `--caching` option that can be specified on the command line, there is also the possibility of using a configuration file in which a user can not only set the default cache expiry lifetime (to be used when the `--caching` option is not used), but also set different cache expiry lifetimes for different cached files, for example:

```
# Refresh FBbt if older than 1 month
fbbt.db = 1m
# Refresh other FlyBase ontologies if older than 2 months
fb*.db = 2m
# Refresh Uberon if older than 2 weeks
uberon.db = 2w
# For any other file, refresh if older than 1 month
default = 1m
```

cmungall (approved):

PR #800, opened 2024-08-19
Bug in calculating `all_by_all_pairwise_similarity`

hrshdhgd:
When `len(ancestor_set) == 0` the jacquard and AIC scores are explicitly set to 0 which is incorrect. 

Thank you @souzadevinicius  for highlighting this!

Also updated `semsimian` version dependency to latest

fixes https://github.com/monarch-initiative/semsimian/issues/133#issuecomment-2286663198

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -5159,46 +5159,46 @@ stats = ["scipy (>=1.3)", "statsmodels (>=0.10)"]
 
 [[package]]
 name = "semsimian"
-version = "0.2.16"
+version = "0.2.18"
 description = "Sematic similarity calculations for ontologies implemented in Rust."
-optional = true
+optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "semsimian-0.2.16-cp310-cp310-macosx_10_12_x86_64.macosx_11_0_arm64.macosx_10_12_universal2.whl", hash = "sha256:b8ad450f4fe8c4f3665e33569d5cdc1a8de9b128a41f3e5dc707a2ba64f16d1a"},
-    {file = "semsimian-0.2.16-cp310-cp310-macosx_10_12_x86_64.whl", hash = "sha256:03065a9fce6b249315db82883164f500476604606d1f0583f9401526bf5bf748"},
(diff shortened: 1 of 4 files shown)


justaddcoffee (approved):

caufieldjh (approved):

caufieldjh:
OK, so what was happening was that the provided nodes didn't have ancestors (at least not in the provided set), so their metrics were set to zero. Should this produce an error and/or should there be an attempt (in oaklib or semsimian) to identify the potential ancestor? It's reasonable to expect that the user should only need to provide IC values for nodes of interest, even if the ancestors are required to calculate pairwise semsim.

justaddcoffee:
Could the 3 of us do a code review on this, maybe tomorrow if we are all around?

> OK, so what was happening was that the provided nodes didn't have ancestors (at least not in the provided set), so their metrics were set to zero.

If I understand the code correctly, we check [here](https://github.com/INCATools/ontology-access-kit/blob/c580d2dde9ece11b1e928accec47fd47700ff675/src/oaklib/implementations/semsimian/semsimian_implementation.py#L224) if the two terms have a common ancestor (which we call `ancestor_set`, but I think is a bad name - it's a tuple of ancestor_id and the IC value of this ancestor). I think the correct behavior is what we do here, which is to set the common ancestor ID in the similarity object if we have it, otherwise set it to OWL_THING or some such. 

I think the remaining question is why so many (all?) of Vinicius's pairs of ontology terms didn't have a common ancestor? This seems like a bug somewhere, either in OAK, semsimian, or Phenio.

cmungall (approved):

caufieldjh:
> I think the remaining question is why so many (all?) of Vinicius's pairs of ontology terms didn't have a common ancestor? This seems like a bug somewhere, either in OAK, semsimian, or Phenio.

They certainly have common ancestors, and in PHENIO the common ancestor of an HP term and an MP term will almost always be a UPHENO term. But somehow if the UPHENO term isn't included in the provided IC map it doesn't get included as an ancestor, *even if it's present in the source ontology* - perhaps because we don't mix user-provided IC maps with calculated ones.

Issue #838, opened 2025-04-09
chore: improve dev-testing ergonomics


sierra-moxon:
- [ ] improve dev control over test outputs (e.g., move `print` to `logging`/`logger`) 
- [ ] improve test run speed (currently takes ~ 38 mins locally to run the full suite)
- [ ] convert unittest to pytest 
- [ ] debug why `make test` ends up hanging after and forces users to cntrl-c:
```
----------------------------------------------------------------------
Ran 503 tests in 2309.364s

OK (skipped=94)
```

PR #839, opened 2025-04-09
Test upgrade of linkml-runtime to 1.9.1-rc1

sierra-moxon:
This also reduces some test verbosity (move print -> logging.info, but the logger is not configured to completion)
adds a short-cut `make lint-fix` target.


tests/test_cli.py
```diff
@@ -87,7 +87,7 @@ def _out(self, path: Optional[str] = TEST_OUT) -> str:
     def test_main_help(self):
         result = self.runner.invoke(main, ["--help"])
         out = result.stdout
-        print("STDERR", result.stderr)
+        logging.info("STDERR", result.stderr)
```
dalito, line 90: Maybe switch to a named logger as suggested here ([tutorial](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial) or [logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html#logging-cookbook)).

To change insert after the import section
```
logger = logging.getLogger(__name__)
```
and then replace all logging.debug, ligging.info etc. by logger.debug, logger.info etc.

linkml-runtime also does it this way ([test_schemaview.py](https://github.com/linkml/linkml-runtime/blob/main/tests/test_utils/test_schemaview.py)).

sierra-moxon: yep!  definitely.  not done with this implementation at all, just first pass at getting the output a bit more grok'able so I can see the breaking errors.

sierra-moxon: I think this refactoring should go in another PR.  The changes here are consistent with the rest of the library and I captured the next step we need to take here: https://github.com/INCATools/ontology-access-kit/issues/838

PR #849, opened 2026-02-20
Pyproject.toml Update to python standard/UV Update For Oak

DnlRKorn:
Standardized pyproject.toml to [PEP 621](https://peps.python.org/pep-0621/). This allows us to use UV to build/run oak. Currently maintaining previous poetry functionality (which all tested to work). Added job to github workflows to have some also run test using UV (had to make minor changes to get all workflows functional). 

Maintained all existing functionality by fixing versions (changes noted with TODO in the pyproject indicating where in code failures were occuring). Only changed 2 lines actual code which had no possible workaround using package versioning.

### DIFF
--- a/.github/workflows/pypi-publish-dry-run.yaml
+++ b/.github/workflows/pypi-publish-dry-run.yaml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build-n-publish-dry-run:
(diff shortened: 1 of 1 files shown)


cmungall (approved):
Lgtm. I don't know if the staged transition to Uv is necessary I favor the cropping the bandaid approach. Cc is particularly good at this. But any path welcome!

PR #900, opened 2026-07-14
Remove curies upper pin and guard against pronto backtrack

### LINKED ISSUE #899: Drop the `curies<0.11` upper cap so oaklib works with modern curies
`oaklib` pins `curies>=0.6.6,<0.11` in `pyproject.toml`, with a TODO noting that curies ≥0.11 makes `TestOboGraphUtils` fail. The cap blocks any environment that needs a newer curies — e.g. the ODK, which wants `curies>=0.13.8` for features added in the 0.11–0.14 line. Because the ceiling sits entirely below `0.13.8`, downstream resolution is impossible.

The cap is a real breakage, not just caution. curies 0.11 (biopragmatics/curies#186, "remove deprecated return behavior") changed `Converter.parse_uri`: for a URI it cannot parse it now returns `None` instead of the old `(None, None)` tuple. `induce_graph_prefix_map` in `obograph_utils.py` unpacked that return directly (`pfx, _ = converter.parse_uri(id)`), so the `None` raised `TypeError` and `TestOboGraphUtils` failed.

The fix is small: handle the `None` return and drop the ceiling. Unpacking via index keeps it working across the whole supported range (parseable URIs still return a `ReferenceTuple`, which indexes fine; older curies still returns `(None, None)`, which is `not None` and indexes to `None`).

While here, floor `pronto>=2.5`. Without a floor, pip is free to backtrack `pronto` into an ancient `0.12.x` release that subclasses the removed `collections.Mapping` and crashes on import under Python ≥3.10 — a trap that shows up once other constraints tighten.

## Examples

Fix:

```diff
-    "curies>=0.6.6,<0.11", #TODO: ... parse_uri is returning None; causing TestOboGraphUtils to fail
-    "pronto",
+    "curies>=0.6.6",
+    "pronto>=2.5",
```

```diff
-        pfx, _ = converter.parse_uri(id)
+        reference = converter.parse_uri(id)
+        pfx = reference[0] if reference is not None else None
         if pfx:
             return pfx
```

Verified on Python 3.12 with `curies==0.14.1`:

```
curies==0.14.1
tests/test_utilities/test_obograph_utils.py .......... [100%]
10 passed
```

cc @cthoyt (just an FYI - will deal with this)

matentzn:
Fixes #899


matentzn:
@cthoyt its a two liner - if you would be so good to review when you have some time 🙏 

Much appreciated.

DnlRKorn (approved):
Lgtm

Issue #901, opened 2026-07-14
Add Python 3.14 to the compatibility matrix


matentzn:
We currently cap Python support below 3.14 and don't test against it. `pyproject.toml` line 8 has `requires-python = ">=3.10,<4,<3.14"`, and the CI matrix in `.github/workflows/main.yaml` line 40 is `[ "3.10", "3.11", "3.12", "3.13" ]`. We should add 3.14 to the test matrix and drop (or bump) the `<3.14` cap.

The real question is dependency support, not the version bump. The `<3.14` cap and the earlier `<3.13` cap from #857 both existed because a transitive dependency (semsimian) forced them. Before relaxing the cap we need to confirm that _semsimian_ ships 3.14 wheels, and check the other dependencies that have already caused friction: eutils (pinned via #852), and curies/pronto (recent commits `296d3b543` and `f70006072` guard against a pronto backtrack and removed a curies upper pin).

If any of those don't support 3.14 yet, that dependency is the blocker and this stays open until it's resolved upstream.

And, of course https://github.com/ontology-tools/py-horned-owl/issues/58 which is implemented but not released yet.

## Related

- #857 — precedent: semsimian forced a down-version below 3.13, the same dependency likely gates 3.14
- #852 — eutils pin, one of the dependencies to re-check against 3.14
- https://github.com/ontology-tools/py-horned-owl/issues/58

DnlRKorn:
I tested oaklib in 3.14 as things stand right now. Presently it doesn't work - the wheel file for py-horned-owl that is on pypi doesn't work (so uv throws an error when it tries to be imported). I made an error for this here https://github.com/ontology-tools/py-horned-owl/issues/60

We can wait for an updated wheel to be uploaded to pypi. However, depending on how pressing this is, we can add the following workaround to the bottom of the pyproject.toml file and everything should pass. 
```
[tool.uv.sources]
py-horned-owl = { git = "https://github.com/ontology-tools/py-horned-owl.git", rev = "main" }
```

matentzn:
😁 the third person that made the issue in the pyhorned owl repo 😁 mine was also closed yesterday so expect the same

DnlRKorn:
Pyhornedowl now works following release 1.4.1 earlier today. Submitting a PR

PR #902, opened 2026-07-15
Adding Python 3.14 support

DnlRKorn:
Adding support and changing our build pipelines to use python 3.14 as requested in #901.


.github/workflows/doc_pages.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
     - name: Set up Python 3.
       uses: actions/setup-python@v6.2.0
       with:
-        python-version: 3.12
+        python-version: 3.14
```
matentzn, line 19: not strictly speaking necessary? - would opt against this is changing release / docs infra, not establishing OAK compatibility

.github/workflows/main.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4.3.0
         with:
-          python-version: "3.12"
+          python-version: "3.14"
```
matentzn, line 29: Not strictly speaking necessary - would opt against this change

.github/workflows/pypi-publish.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v6.2.0
       with:
-        python-version: 3.12
+        python-version: 3.14
```
matentzn, line 28: Again not strictly speaking necessary

pyproject.toml
```diff
@@ -5,7 +5,7 @@ description = "Ontology Access Kit: Python library for common ontology operation
 authors = [
     {name="Chris Mungall", email="cjm@berkeleybop.org"}
 ]
-requires-python = ">=3.10,<4,<3.14"
+requires-python = ">=3.10,<4,<3.15"
```
matentzn, line 8: Why do we need the upper bound at all?

.github/workflows/main.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4.3.0
```
matentzn, line 27: actions/setup-python@v4.3.0 - others are on v6.2.0.. worth updating?

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "pysolr>=3.9.0,<4",
     "eutils>=0.6.0", #TODO: eutils==0.6.1 is only created for python>=3.12. It's forcing us to use setuptools. We need to go into the innards of the code and replace this dependency with an api call - GHIssue 852
     "setuptools<82", #We need this for eutils==0.6.0    
+    "lxml>5.4.0", #Need this for eutils
```
matentzn:
```suggestion
    "lxml>5.4.0", # floor for 3.14 wheels; needed for eutils `>=0.6.0`
```

pyproject.toml
```diff
@@ -156,3 +157,4 @@ target-version = "py310"
 [tool.ruff.lint.mccabe]
 # Unlike Flake8, default to a complexity level of 10.
 max-complexity = 10
+
```
matentzn, line 162: harmless but spurious

matentzn (commented):
AWESOME - 🎉 🎉 

Some voluntary comments to deal with

pyproject.toml
matentzn: Do we know that 3.15 will fail?

DnlRKorn: I think, we should be defensive about allowing unknown python versions run without first testing them. Otherwise a bunch of pipelines could randomly break the day python releases 3.15 and automatically changes the uv builds.

DnlRKorn:
@matentzn I'm of the opinion that if we're trying to support the most recent version of python we should stake all of our build processes onto the most up-to-date version of python. "Rip off the band-aid" as Chris has put it before.

My rationale for this is that we're unlikely to touch these workflows and code again, so having these processes be on an older version of python leaves it as an unexplored space for finding bugs/issues.

.github/workflows/doc_pages.yaml
DnlRKorn: Can confirm this ran without issue, so doesn't pose a risk to failing build

https://github.com/INCATools/ontology-access-kit/actions/runs/29427519795

pyproject.toml
matentzn: https://github.com/linkml/linkml/pull/3675#pullrequestreview-4570277821

matentzn:
> @matentzn I'm of the opinion that if we're trying to support the most recent version of python we should stake all of our build processes onto the most up-to-date version of python. "Rip off the band-aid" as Chris has put it before.

Not against this but more important is separation of concerns IMO; I cannot confidently approve something _quickly_ if I cant be certain of the consequences. But I agree with the general sentiment, would just like to move these to a different PR so we can merge in stages (and add some more confident reviewers).

That said, this is only a mild push back. If you still believe we should funnel in one PR, lemme know and I will approve. I assume you will volunteer fixing issues unrelated to the matrix change should they arrive :P

DnlRKorn:
I'll take responsibility of bug hunting/rollbacks should things start falling apart

.github/workflows/main.yaml
matentzn: https://github.com/INCATools/ontology-access-kit/issues/904

pyproject.toml
matentzn: https://github.com/INCATools/ontology-access-kit/issues/905

matentzn (approved):

pyproject.toml
DnlRKorn: I think this makes argument sense and applies to minor versions, but I'd disagree with not being defensive with untested major versions.

Python made a lot of breaking changes when moving from 3.12->3.13 and 3.13->3.14; they weren't unreasonable changes, but required a lift on our part to make our code compliant.

Issue #64, opened 2022-05-04
Ensure curies are always curies


cmungall:
many of the datamodels force the use of CURIEs

sometimes applying naive uri-to-curie yields things like

`uberon/core#BRAIN:NAME:ABV`

not clear if this is due to some kind of roundtrip through obo but either way the code should be robust and always yield CURIEs where these are expected

PR #911, opened 2026-09-16
Fix the datamodels toc and generate the value-set-configuration docs

TankLin23:
The datamodels section of the published documentation does not match what the
build actually produces. Three separate causes, all silent because
`sphinx-build` is invoked without `-W`, so a toctree that fails to resolve only
produces a warning.

### 1. `gendoc-vsc` never runs

`gendoc-vsc` is defined in the Makefile, but it was never added to
`run-all-gendoc-cmds` — which is the only target the docs workflow runs:

```yaml
# .github/workflows/doc_pages.yaml
- name: Run all gendoc commands
  run: make run-all-gendoc-cmds
```

So `docs/datamodels/value-set-configuration/` is never created, and the page
that the toc expects is missing from the site:

```
$ curl -s -o /dev/null -w "%{http_code}\n" -L \
    https://incatools.github.io/ontology-access-kit/datamodels/value-set-configuration/index.html
404
```

This adds it to the aggregate target.

### 2. Two toc entries do not resolve

`docs/datamodels/index.rst` disagrees with the directory names that `gen-doc`
is told to write into:

| toc entry | `gen-doc -d` writes to | result |
|---|---|---|
| `item-lists/index` | `docs/datamodels/item-list` | 404 |
| `value_set_configuration/index` | `docs/datamodels/value-set-configuration` | 404 |

The item list datamodel is generated and published — it is reachable at
`/datamodels/item-list/index.html` (200) — but nothing in the navigation links
to it, because the entry is spelled `item-lists`. Same story for the
underscore in the value set configuration entry. Both are corrected here.

### 3. Two pages exist but are unreachable

`class-enrichment` (produced by `gendoc-ce`) and `semsql` (a hand-written page
next to the other datamodel pages) both exist on the site but were not listed
in the toc, so they were orphaned. They are added.

### One removal

The `fhir/index` entry is dropped. It does not correspond to anything: there is
no `gendoc-fhir` target for `fhir.yaml`, and no `docs/datamodels/fhir/`
directory, so the entry could never resolve. If fhir datamodel docs are wanted,
the entry should come back together with a target and a page — happy to do that
as a follow-up if that is the intention.

### Verification

Parsing the aggregate target, the `-d` output directories and the toc gives:

| | before | after |
|---|---|---|
| toc entries pointing at nothing | `fhir/index`, `item-lists/index`, `value_set_configuration/index` | none |
| generated but not in the navigation | `class-enrichment`, `item-list` | none |
| targets defined but not in the aggregate | `gendoc-vsc` | none |

No code is touched; the change only affects which documentation pages are
generated and linked.


matentzn:
@TankLin23 thanks for your contributions. Can you tell us a bit who you are and why you are interested in contributing to ODK? Its rare we get that many drive by PRs (and they look sane), but I want to make sure you are not a bot before moving on (you dont have a long standing github history to tell.

TankLin23:
Hi Nico— thanks for asking, and it's a fair question.

I'm Tank (@TankLin23), an AI engineer based in mainland China. I use this toolchain in real work at my company, which is how I ended up reading the ODK and OAK source rather than just the docs. Contributing back to the parts I rely on seemed more useful than filing issues I could not act on. The account is from 2019, for what it's worth — it has just been quiet.

I'll be upfront about the volume too: I used an AI assistant to scan the repos and to run the checks written up in the descriptions, then went through the results myself before opening anything. So the burst is one person plus an assistant working through four repos I already had checked out, not a bot crawling GitHub. The verification in the descriptions is real output from those checks — live `curl` against the deployed sites, patches applied to fresh clones — rather than boilerplate.

Two things that might say more than my word:

- I closed #64 in linkml-reference-validator myself once I found that the modular full-text fetching plan already keeps that code path. I would rather withdraw a PR than leave a stale, conflicting branch around.
- Each of these is small and independent, and none of them overlap with the others. I can explain what each change does and why it is right — and if something comes up that I cannot answer confidently, I would rather close the PR than improvise.

If the volume is unwelcome, tell me and I will close whatever you would rather not deal with. And if you have a policy on AI-assisted contributions, I will follow it.

Happy to answer anything else — about me, or about the changes.

matentzn:
@TankLin23 

thank you for the answer; of course, its not possible from that to decide purely based on that if you are a human in our day and age, but lets say ok :) I will review your PRs soon, dont close them. Note that we are aware that especially when it comes to the docs, much larger refactorings are needed, but I am not opposed to folding in minor improvements once I convince myself they are that.

TankLin23:
Thanks,  Nico @matentzn .I Won't close any of them, and I won't open anything new until you've had a chance to look at them.

On the docs — fair, and it lines up with what I ran into while reading through the repo. Point fixes won't get those pages where they need to be, and I'd rather spend the effort on the restructuring you have in mind than keep picking at the edges.

If a piece of that is up for grabs, I'd be glad to take it. Same goes for anything else that's been sitting around because nobody had the time: point me at it and I'll work through it properly. Slow is fine, and so is being told my first attempt is wrong.

That's really why I'm here. I use this toolchain in my day job. What goes back into ODK or OAK goes back into something I depend on, so I'd rather put my time into the work you're leading than into drive-by patches.

Happy to open an issue first if you'd rather have a plan before anyone touches code.

